PR #2, opened 2021-04-03
chore: remove unnecessary arc

driftluo:
1. Global static variables do not need to be wrapped with Arc again
2. Fix module change cause ci fail

By the way, the project doesn’t seem to be formatted with [rustfmt](https://github.com/rust-lang/rustfmt), I’m not sure if you’re willing to do this, so I didn’t use it rashly either


chrislearn:
Thanks, I will use rustfmt

PR #3, opened 2021-04-11
fix: allow single feature on extra

driftluo:
In the current code, if all features are not allowed, a single feature cannot be used


chrislearn:
👍

PR #20, opened 2021-04-22
show summary and detail for anyhow::Error

siwilizhao:


chrislearn:
Is there security issue when display error details to any user? Or we just display error details in debug mode?

driftluo:
> Is there security issue when display error details to any user? Or we just display error details in debug mode?

This is actually a difficult decision. I personally recommend that users use a newtype wrapper anyhow error to implement Writer, but it requires more operations from the user to determine what can be exposed to the outside world.

PR #28, opened 2021-05-11
star name regex supported

chrislearn:


driftluo (commented):
I think it is necessary to add instructions on the special paths `*xxx` and `hex` in the readme

PR #31, opened 2021-05-13
Fix dir list streaming bug

chrislearn:
https://github.com/salvo-rs/salvo/issues/29#issuecomment-840434616

### DIFF
--- a/extra/src/compression/mod.rs
+++ b/extra/src/compression/mod.rs
@@ -71,23 +71,47 @@ impl Handler for CompressionHandler {
         {
             return;
         }
-        let body = res.take_body().unwrap();
-        if let Body::Empty = body {
-            return;
-        }
-        let body = body.map(|item| item.map_err(|_| std::io::ErrorKind::Other));
-        match self.algo {
-            CompressionAlgo::Gzip => {
-                let stream = ReaderStream::new(GzipEncoder::new(StreamReader::new(body)));
-                res.streaming(stream);
-            }
-            CompressionAlgo::Deflate => {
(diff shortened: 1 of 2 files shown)


biluohc:
感谢，

不过也许
```
pub use dir::{StaticDir, Options};
```
比 
```
pub mod dir;
pub mod fs;

pub use dir::StaticDir;
pub use fs::StaticFile;
```
好一些？

chrislearn:
Option s 直接拉出来意义好像不那么明确

biluohc:
主要是options这个名字吧，，模块都公开的话StaticDir/StaticFile就会有两个路径可用了，，虽然也不是什么比较大的问题

extra/src/compression/mod.rs
```diff
@@ -71,23 +71,47 @@ impl Handler for CompressionHandler {
...
-                let stream = ReaderStream::new(BrotliEncoder::new(StreamReader::new(body)));
-                res.streaming(stream);
+        if let Some(body) = res.take_body() {
+            match body {
+                Body::Empty => {
+                    return;
+                }
+                Body::Bytes(body) => {
+                    let reader =
+                        StreamReader::new(tokio_stream::iter(vec![Result::<_, std::io::Error>::Ok(body)]));
```
driftluo, line 81: use `once` instead, don't need vec

driftluo (commented):
I think you can test the compression ratio and set the lower limit of compression for users. Small data compression is an idling CPU, which has no practical meaning.

biluohc:
压缩确实是，太小的body压缩后会比原始的还大

Issue #414, opened 2023-09-13
Uuid is not supported in oapi::endpoint macro


Roba1993:
**Describe the bug**
ToSchema seems not to be implemented completly for Uuid.

**To Reproduce**
My `Cargo.toml` configuration:
```toml
[dependencies]
salvo = { version = "0.55", features = ["rustls", "serve-static", "affix", "websocket", "compression", "anyhow", "session"] }
salvo-oapi = { version = "0.55", features = ["swagger-ui", "rapidoc", "uuid", "time"] }
uuid = { version = "1.2", features = ["v4", "serde"] }
```

How I use the `#handle` macro:
```rust
#[endpoint]
pub async fn get_record_rest(
    env_uuid: HeaderParam<Uuid, true>,
) -> Json<String> {
    Json("test".to_string())
}
```

**Expected behavior**
The Uuid typ should be supported.

chrislearn:
Have you enable serde feature in Cargo.toml?

```toml
[dependencies]
uuid = {  workspace = true, features = ["serde"] }
```

Roba1993:
Yes I have, sorry I have not put all dependencies into the example...

chrislearn:
fixed in version: v0.55.3

PR #415, opened 2023-09-14
fix: impl `ToSchema` for `Uuid` and other types #414

chrislearn:


Roba1993:
Works perfectly. Thank you 👍

Issue #454, opened 2023-10-16
add support for https://github.com/scalar/scalar as alternative for swagger ui


prabirshrestha:
I'm primarily interested in using the feature which allows be to copy as curl or other programming languages.

https://github.com/scalar/scalar
https://news.ycombinator.com/item?id=37871577

chrislearn:
Can you give a PR?

PR #458, opened 2023-10-18
Add support for scalar as alternative for swagger ui

chrislearn:
#454

### DIFF
--- a/crates/oapi/Cargo.toml
+++ b/crates/oapi/Cargo.toml
@@ -14,8 +14,9 @@ authors = ["Juha Kukkonen <juha7kukkonen@gmail.com>", "Chrislearn Young <chrisle
 
 [features]
 default = []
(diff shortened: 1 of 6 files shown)


prabirshrestha:
Awesome. Thanks for adding this. Curious what it would take to share the files withing the crate so there is are no dependencies with the public internet. This could be solved later with the following api. 

```rust
Scalar::new("/api-doc/openapi.json").with_script_path("/public/@scalar/api-reference")
```

chrislearn:
> Awesome. Thanks for adding this. Curious what it would take to share the files withing the crate so there is are no dependencies with the public internet. This could be solved later with the following api.
> 
> ```rust
> Scalar::new("/api-doc/openapi.json").with_script_path("/public/@scalar/api-reference")
> ```

Done: https://github.com/salvo-rs/salvo/pull/461

prabirshrestha:
Thanks!

PR #500, opened 2023-11-21
Resolve Mismatched Types Error in Macro Expansion

fankaiLiu:
- The error originated from the attribute macro endpoint and was primarily due to the naming conflict between user-defined variables and those generated by the macro.
like this:
```rust
#[endpoint(tags("todos"), status_codes(201, 409))]
pub async fn create_todo(req: JsonBody<Todo>) -> Result<StatusCode, StatusError> {
    tracing::debug!(todo = ?req, "create todo");

    let mut vec = STORE.lock().await;

    for todo in vec.iter() {
        if todo.id == req.id {
            tracing::debug!(id = ?req.id, "todo already exists");
            return Err(StatusError::bad_request().brief("todo already exists"));
        }
    }

    vec.push(req.into_inner());
    Ok(StatusCode::CREATED)
}
//The code he generates is
fn handle<'life0, 'life1, 'life2, 'life3, 'life4, 'async_trait>(
&'life0 self,
//here
req: &'life1 mut salvo::Request,
depot: &'life2 mut salvo::Depot,
res: &'life3 mut salvo::Response,
ctrl: &'life4 mut salvo::FlowCtrl,
) -> ::core::pin::Pin<Box<dyn ::core::future::Future<Output = ()> + ::core::marker::Send + 'async_trait,>,>
where
'life0: 'async_trait,
'life1: 'async_trait,
'life2: 'async_trait,
'life3: 'async_trait,
'life4: 'async_trait,
Self: 'async_trait,
{
Box::pin(async move {
let __self = self;
let () = {
//here
let req: JsonBody<Todo> = match <JsonBody<Todo,> as salvo::Extractible>::extract_with_arg(req, "req").await
...
}
```
- the macro generated variables with common names such as req and request. If a user happened to use these same names in their code, it would lead to a name collision and subsequently a type mismatch error, as the macro-generated variable was of a different type than the user-defined one.
- To solve this issue, I updated the macro to generate variables with unique names, specifically by adding a __macro_generated_ prefix to the generated variable names.


fankaiLiu:
Looks like some of the tests didn't pass, I'll check them out

PR #575, opened 2023-12-19
Read all cookie headers to fill `CookieJar`

### LINKED ISSUE #573: Possible bug: when multiple "cookie" fileds present in Request.headers, instead of separated by ";" under a single "cookie" key, failed to fetch all of them into Request.cookies
**Describe the bug**
I see this really odd behavior: when testing my site on my local desktop, without https enabled, I noticed if there're two cookies: one is my session.id, another one is added by a script (such as google analytics) or manually added by hands (google chrome->developer tools -> application -> Cookies),  the requests are sent with two cookies under a single cookie key "cookie" when print out req:
```rust
"cookie": "test=test; session.id=SQotRLwglxKyE+Q8UxUmSvGGD7yyJqxD"
```
I guess this is "correct" behavior since the session can be fetched without problem and works as expected. But when the same thing is tested in production with https enabled, I noticed the two cookies are now in separate keys (yes, the HeaderMap's key is not unique), something like this:
```rust
        "cookie": "test=test",
        "cookie": "session.id=lfbGFuZwQAAAAAAAAAInpoIg%3D%3D"
```
If I print out req.cookies(), it only contains the first cookie:
```rust
req.cookies() = CookieJar {
    original_cookies: {
        DeltaCookie {
            cookie: Cookie {
                cookie_string: Some(
                    "test=test",
                ),
                name: Indexed(
                    0,
                    4,
                ),
                value: Indexed(
                    5,
                    9,
                ),
                expires: None,
                max_age: None,
                domain: None,
                path: None,
                secure: None,
                http_only: None,
                same_site: None,
            },
            removed: false,
        },
    },
    delta_cookies: {},
}
```
As a result, the session fetched from depot will be None. I think this is the relevant code:
[request.rs](https://github.com/salvo-rs/salvo/blob/1d1940e9fa8cf05e886e26c927ec17f5c2145f8f/crates/core/src/http/request.rs#L142C1-L142C1)
```rust
        // Set the request cookies, if they exist.
        #[cfg(feature = "cookie")]
        let cookies = if let Some(header) = headers.get("Cookie") {
            let mut cookie_jar = CookieJar::new();
            if let Ok(header) = header.to_str() {
                for cookie_str in header.split(';').map(|s| s.trim()) {
                    if let Ok(cookie) = Cookie::parse_encoded(cookie_str).map(|c| c.into_owned()) {
                        cookie_jar.add_original(cookie);
                    }
                }
            }
            cookie_jar
        } else {
            CookieJar::new()
        };
```
**To Reproduce**
As described above, I can only reproduce this with acme / https enabled.

**Expected behavior**
all cookies should be put into Request.cookies, thus session can be fetched from CookieStore as expected.

**Desktop (please complete the following information):**
 - OS: ubuntu 22.04.3 LTS
 - Browser: google chrome
 - Version: Version 117.0.5938.92 (Official Build) (64-bit)

chrislearn:
Close #573

### DIFF
--- a/crates/core/src/http/request.rs
+++ b/crates/core/src/http/request.rs
@@ -152,7 +152,6 @@ impl Request {
                 }
             }
             cookie_jar
-
         };
 
         Request {
--- a/crates/core/src/http/request.rs
+++ b/crates/core/src/http/request.rs
@@ -8,7 +8,7 @@ use std::sync::Arc;
 use bytes::Bytes;
 #[cfg(feature = "cookie")]
 use cookie::{Cookie, CookieJar};
-use http::header::{AsHeaderName, HeaderMap, HeaderValue, IntoHeaderName, CONTENT_TYPE};
+use http::header::{AsHeaderName, HeaderMap, HeaderValue, IntoHeaderName, CONTENT_TYPE, COOKIE};
 use http::method::Method;
 pub use http::request::Parts;
 use http::uri::{Scheme, Uri};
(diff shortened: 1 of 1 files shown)


chrislearn:
I fixed it. please check gihub main branch

Issue #617, opened 2024-01-10
[Bug]: `ToSchema` will return an error when set `#[serde(flatten)]` on a field


TheAwiteb:
### Describe the bug
When set `#[serde(flatten)]` on a field, `ToSchema` will return an error saying `failed to resolve: use of undeclared crate or module 'utoipa'
use of undeclared crate or module 'utoipa'`

### To Reproduce
The code is as follows:
```rust

#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
#[salvo(schema(symbol = "UserSchema", example = json!(UserSchema::default())))]
pub struct UserSchema {
    // ...
}

#[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
#[salvo(schema(symbol = "NewUserResponseSchema", example = json!(NewUserResponseSchema::default())))]
pub struct NewUserResponseSchema {
    #[serde(flatten)]
    pub user: UserSchema,
    pub jwt: String,
}
```

When I remove `#[serde(flatten)]`, it works well.

### Environment
- Rustc version: `rustc 1.75.0 (82e1608df 2023-12-21)` (stablel)
- Salvo version: `rev = 444af08b` (latest) (features: `["logging", "affix", "oapi", "jwt-auth"]`)
- OS: Pop!_OS 22.04 LTS x86_64 

## Expected behavior
The `UserSchema` should be flatten into `NewUserResponseSchema` and the `ToSchema` should work well.

chrislearn:
Should be fixe by #618

TheAwiteb:
A new error has appeared after this [PR#618](https://github.com/salvo-rs/salvo/pull/618)

```
error: expected identifier, found `:`
  --> src/schemas/jwt.rs:63:48
   |
63 | #[derive(Serialize, Deserialize, Clone, Debug, ToSchema)]
   |                                                ^^^^^^^^
   |                                                |
   |                                                expected identifier
   |                                                in this derive macro expansion
   |
  ::: /home/awiteb/.cargo/git/checkouts/salvo-7f211b7208170219/063b292/crates/oapi-macros/src/lib.rs:73:1
   |
73 | #[proc_macro_error]
   | ------------------- in this expansion of `#[derive(ToSchema)]`
   |
   = note: type ascription syntax has been removed, see issue #101728 <https://github.com/rust-lang/rust/issues/101728>

error: proc-macro derive produced unparsable tokens
  --> src/schemas/jwt.rs:63:48
   |
[...]

PR #618, opened 2024-01-10
Fix extract attribute conflict with serde

chrislearn:


TheAwiteb:
Fix https://github.com/salvo-rs/salvo/issues/617

TheAwiteb:
Ok, I can't link it. Can you link it?

PR #686, opened 2024-02-15
Add cfg_attr(docsrs) to compression crate

jkelleyrtp:
The compression crate uses cfg_attr(docsrs) to use nightly docs features but does not actually declare those nightly features, breaking all downstream consumers of salvo.

This fixes that by enabling the appropriate nightly feature.

### DIFF
--- a/crates/compression/src/lib.rs
+++ b/crates/compression/src/lib.rs
@@ -1,6 +1,9 @@
+#![cfg_attr(docsrs, feature(doc_cfg))]
+
 //! Compression middleware for for Savlo web server framework.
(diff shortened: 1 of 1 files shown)


jkelleyrtp:
@chrislearn Thanks! We use salvo in Dioxus-fullstack, and with the currently published version of salvo, can't run nightly docs against our packages. Do you think you can do a minor release with this change incorporated? This would let us keep support for salvo in Dioxus 0.5.

chrislearn:
new version 0.66.0 released.

PR #730, opened 2024-03-26
Add reqwest client to proxy middleware

zhangkai803:


zhangkai803:
This PR was basically a copy from older version with testcases added.

Issue #867, opened 2024-08-14
Rename hoop


clarnx:
**Is your feature request related to a problem? Please describe.**
A quick Google search of the word `hoop` does not describe anything related to middleware.

**Describe the solution you'd like**
A clear and concise description of an alternative name for `hoop` will be ideal since most developers are familiar with the word `middleware` in other frameworks or programming languages. Salvo is simple enough to attract newer developers learning Rust and coming from Laravel, Express.js, and Django backgrounds. Laravel, Express.js, and Django all use the word middleware.

**Describe alternatives you've considered**
Some suggestion in my opinion as an alternative name to hoop can be `middleware`. 

```rust
Router::new()
    .middleware(check_authed)
    .path("writers")
```

chrislearn:
Hoop is a very suitable word. I don’t like the word middleware because it is too long.

clarnx:
> Hoop is a very suitable word. I don’t like the word middleware because it is too long.

Thanks for the response. @chrislearn Another way know if the community wants `hoop` or `middleware` is to create a poll for the community to choose to keep hoop or replace it with `middleware`.

If you still want to keep hoop, another alternative to achieving this is that `hoop` still remains and then you add `middleware` which functions just like `hoop` for those who want to use it to make their code more relatable to other developers on a project.

astoring:
My personal opinion is that changing the name is useless.
Middleware is a very basic function. Even beginners only need to look at the example or the documentation to know that hoop is middleware.

Adding another middleware method will inevitably make people wonder what the relationship between the two functions `middleware` and `hoop` is, which will confuse users. If you really want to add it, you must mark it as a deprecated function.

TheAwiteb:
I don't think there is any need for this, if you really need to write middleware instead of hoop, create a trait extension like this

<details>

<summary>The code</summary>

```rust
use salvo::logging::Logger;
use salvo::prelude::*;

mod traits_ext {
    use std::sync::Arc;

    use salvo::{Handler, Router};

    #[easy_ext::ext]
    pub impl Router {
        fn middleware(mut self, m: impl Handler) -> Self {
            self.hoops.push(Arc::new(m));
            self
        }
    }
}

use traits_ext::*;

#[handler]
async fn hello() -> &'static str {
    "Hello World"
}

#[tokio::main]
async fn main() {
    tracing_subscriber::fmt().init();

    let acceptor = TcpListener::new("127.0.0.1:5800").bind().await;
[...]

PR #870, opened 2024-08-14
Added middleware method to catcher.rs

clarnx:
#867 
Added an implementation method for Catcher in catcher.rs that serves as an alternative to the hoop method.


clarnx (commented):
fixed typo

Issue #945, opened 2024-10-04
OpenAPI macro It may not support the generic parameters.


andeya:
**Describe the bug**
`"handle < U >"` is not a valid identifier.

**To Reproduce**
```rust
#[derive(Debug, Clone)]
pub struct handle<U>(U);

#[salvo::oapi::endpoint]
impl<U> handle<U> {
    async fn handle(&self, _req: &mut Request) -> String {
        String::new()
    }
}
```

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac
 - Version 0.73.0

**Additional context**
1. `#[salvo::handler]` is ok.
2. `pub struct handle(i8)` is also ok.

PR #952, opened 2024-10-10
fix(oapi-macros): Support the generic parameters. #945

andeya:


andeya:
This PR only solves part of the problem. It is temporarily closed. Later, the fix will continue on the [#fix-oapi](https://github.com/salvo-rs/salvo/tree/fix-oapi) branch.

PR #1018, opened 2025-01-05
Change routing path parameter syntax style

chrislearn:
The parameter matching pattern is changed to `{{param}}`, because most web frameworks use this pattern, and the OpenAPI standard also uses this pattern.

If a regular expression is required, the previous is similar to `<param:/\d+/>`. The '/' here is easily confused with the '/' parameter in the path. Now it is changed to the `{param|\d+}` pattern.


EluvK:
It took me about five minutes to figure out why the router didn't hit the handler after updating this...🥲

Issue #1030, opened 2025-01-23
matched-path bug


logicin:
**Describe the bug**
matched-path may have bug.
for example:  
original:  /open/a/b/c，but the "mathced-path" is :  /open/open/open/a/b/c。 
may has many start with "/open/xxx" path in the router defined 。

**To Reproduce**

**Expected behavior**

**Screenshots**
![Image](https://github.com/user-attachments/assets/561e7d72-7246-4cd7-b613-b0990e8a74cd)

**Desktop (please complete the following information):**
 - OS: [MacOS]
 - Browser [chrome]
 - Version [22]

**Additional context**

logicin:
PR.      https://github.com/salvo-rs/salvo/pull/1031

PR #1031, opened 2025-01-23
fix: fix matched-path bug (#1030)

logicin:
fix matched-path bug (#1030)


chrislearn:
If you have time, you can help add some unit tests for this feature.

logicin:
need I close this PR first ? @chrislearn

PR #1045, opened 2025-02-02
fix: bug with colliding headers in `salvo-proxy`

markcda:
I have a Flask service that inserts three `Set-Cookie` headers during the authentication process. But `salvo-proxy` returns only one of these headers for now. So, I think, we need to make this correction.

### DIFF
--- a/crates/proxy/src/lib.rs
+++ b/crates/proxy/src/lib.rs
@@ -320,11 +320,13 @@ where
                             body,
                         ) = response.into_parts();
                         res.status_code(status);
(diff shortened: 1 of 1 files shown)


markcda:
Not fixed yet (tested)

markcda:
If `headers` aren't gonna change, maybe, consider unsafe transmute operation instead of using `value.to_owned()`?

```rust
for name in headers.keys() {
    for value in headers.get_all(name) {
        let owned_value = unsafe { std::mem::transmute(value) };
        res.headers.append(name, owned_value);
    }
}
```

PR #1058, opened 2025-02-16
style(tracing): 

gustavbrlty:
add a newline with the character ⮡ in order to be able to get the server output in a smaller width terminal

example of what it gives : 

image


chrislearn:
We sholud not control log style by using `println`.

PR #1228, opened 2025-11-12
feat: introduce a Proxy::host_header_getter method to alter the way host header is propagated to the upstream

r00t3g:
This PR implements additional getter to obtain a host header for the proxied request along with three predefined host header acquisition methods:
* default_host_header_getter - the one currently used in Salvo and which is used by default
* rfc2616_host_header_getter - the one that amends the port from the target URI if it differs from the standard HTTP/HTTPS ports, which in turn replicates the default NGINX behavior (see https://www.rfc-editor.org/rfc/rfc2616.html#section-14.23)
* preserve_original_host_header_getter - the one that propagates the original request's host header to the proxied request

These changes would facilitate migration to standalone HTTP-self-handling Salvo based apps from apps with NGINX ingress via proxy_pass or any other ingress proxy like Envoy, Traefik or any other. 

Unfortunately I haven't yet been able to test the changes on production loads as they are intended for currently in-development version and had to rely mostly on the tests.

### DIFF
--- a/crates/proxy/src/lib.rs
+++ b/crates/proxy/src/lib.rs
@@ -43,11 +43,11 @@ use std::error::Error as StdError;
 use std::fmt::{self, Debug, Formatter};
 
 use hyper::upgrade::OnUpgrade;
-use percent_encoding::{CONTROLS, utf8_percent_encode};
-use salvo_core::http::header::{CONNECTION, HOST, HeaderMap, HeaderName, HeaderValue, UPGRADE};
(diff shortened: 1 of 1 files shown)


chrislearn:
ci failed

r00t3g:
@chrislearn my bad, used the wrong version when running local checks. :facepalm: Fixed. 

As for typos check, this hits extract/metadata.rs in the core crate which was not changed, and I am not risking to fix it. I suppose it might be somehow related to crate-ci/typos update to v1.39.0 since its version is not fixed in the action.

PR #1230, opened 2025-11-12
feat: Implement client IP forwarding support

r00t3g:
Implemented client IP forwarding in Proxy via X-Forwarded-For HTTP header.

Please note, that the `local-ip-address` crate was added to dependencies to provide system IP address detection.


chrislearn:
Thanks

PR #1232, opened 2025-11-16
feat: add support for multiple and wildcard SNI names in keycerts

cnlancehu:
This PR updates the TLS certificate selection logic to support multiple domain names per certificate and to add wildcard domain matching.

However, to keep the implementation straightforward, it introduces a **breaking change** in [`RustlsConfig.keycerts`](https://github.com/salvo-rs/salvo/commit/2818edbba53ff82977ec79f9f17cd83b66089844#diff-2a04fa795b3bafcb6be17ea0039263ff489dd718d584185bfe78047f9ef806baL165-R165).

Because this change affects the public API and may not represent the ideal long-term design, I hope you will consider refining or redesigning this functionality in a more complete way.

Thank you — this feature is something I urgently need.

### DIFF
--- a/crates/core/src/conn/rustls/config.rs
+++ b/crates/core/src/conn/rustls/config.rs
@@ -15,7 +15,7 @@ use tokio_rustls::rustls::sign::CertifiedKey;
 
 pub use tokio_rustls::rustls::server::ServerConfig;
 
-use crate::conn::IntoConfigStream;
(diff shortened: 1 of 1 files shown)


chrislearn:
keycerts is HashMap, if you set it as some like:

```json
{"a.domain.com": "keycert", "b.domain.com": "keycert"}
```
and then use fallback for wildcard domain, does this meet your needs?

Issue #1122, opened 2025-04-25
HTTP/3 Response stream not closed


Deniskore:
Reproducible minimal example:

```rust
use salvo::conn::rustls::{Keycert, RustlsConfig};
use salvo::prelude::*;
use std::net::SocketAddr;

pub struct Http3Server {
    addr: SocketAddr,
    rustls_config: RustlsConfig,
}

impl Http3Server {
    pub fn new(addr: SocketAddr, cert: Vec<u8>, key: Vec<u8>) -> Self {
        let rustls_config = RustlsConfig::new(Keycert::new().cert(cert).key(key));
        Self { addr, rustls_config }
    }

    pub async fn run(self) {
        let router = Router::new()
            .push(Router::with_path("/hello").get(hello))
            .push(Router::with_path("/").get(index));

        let tcp_listener = TcpListener::new(self.addr).rustls(self.rustls_config.clone());

        let quinn_listener = QuinnListener::new(
            self.rustls_config.build_quinn_config().unwrap(),
            self.addr,
        );

        let acceptor = quinn_listener.join(tcp_listener).bind().await;

        Server::new(acceptor).serve(router).await;
    }
}

#[handler]
async fn hello() -> &'static str {
    "Hello from /hello"
}

#[handler]
async fn index() -> &'static str {
    "Welcome to /"
}

#[tokio::main]
async fn main() {
    // openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -days 365 -nodes
    tracing_subscriber::fmt().init();
    let cert = include_bytes!("../cert.pem").to_vec();
    let key = include_bytes!("../key.pem").to_vec();
    let addr: SocketAddr = "0.0.0.0:4443".parse().unwrap();
    let server = Http3Server::new(addr, cert, key);
    server.run().await;
}
```

Client to check the stream ended flag:

```python
import asyncio
import ssl
from aioquic.asyncio import connect
from aioquic.asyncio.protocol import QuicConnectionProtocol
from aioquic.h3.connection import H3_ALPN, H3Connection
from aioquic.h3.events import HeadersReceived, DataReceived
from aioquic.quic.configuration import QuicConfiguration
from aioquic.quic.events import QuicEvent

class HttpClient(QuicConnectionProtocol):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._http = H3Connection(self._quic)
        self._stream_ended = asyncio.Event()

[...]

PR #1259, opened 2025-12-27
Close HTTP/3 Response stream properly

cnlancehu:
This might be a possible fix for #1122 

The main change is the addition of a `Drop` implementation for the `H3ReqBody` struct, ensuring that the HTTP/3 stream is properly closed when the body is dropped. 

It works well and curl no longer produce error like `HTTP/3 stream 0 reset by server`

image


chrislearn:
Thanks

PR #1268, opened 2026-01-02
feat: add depot for extract

Caisin:


chrislearn:
Don't forget change examples to make compiler happy!

chrislearn:
Honestly, I'm not sure if we should merge this PR. In fact, depot wasn't used in the extract phase. 
There were users who asked whether it's possible to support extracting data from depot, and I think we can discuss that.

Issue #1281, opened 2026-01-09
Multiple query conditions failed to be aggregated and parsed.


driftluo:
**Describe the bug**

code：
```rust
#[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
pub enum DBState {
    #[serde(alias = "open")]
    Open,
    #[serde(alias = "closed_waiting_onchain_settlement")]
    ClosedWaitingOnchainSettlement,
    #[serde(alias = "closed_cooperative")]
    ClosedCooperative,
    #[serde(alias = "closed_uncooperative")]
    ClosedUncooperative,
}

#[derive(Debug, Serialize, Deserialize, Clone)]
#[serde(untagged)]
pub enum State {
    Single(DBState),
    Multiple(Vec<DBState>),
}

#[derive(Debug, Extractible, Serialize, Deserialize)]
#[salvo(extract(default_source(from = "query")))]
pub(crate) struct ByStateParams {
    pub(crate) state: State,
}

#[handler]
pub async fn channel_count_by_state(
    req: &mut Request,
    _res: &mut Response,
) -> Result<String, salvo::Error> {
    let params = req.extract::<ByStateParams>().await?;
    assert_eq!(params.state, State::Multiple(vec![DBState::Open, DBState::ClosedCooperative]));
    Ok(String::new())
}
```

```bash
curl -X GET \
  -H "Accept: application/json" \
  "http://0.0.0.0:8000/group_by_state?state=open&state=closed_cooperative"
```
However, it cannot correctly parse and aggregate multiple states of the query conditions; instead, it always only gets the first value.

After debugging, I found that the problem might be caused by using `get` instead of `get_vec` during parsing:
```rust
#[handler]
pub async fn channel_count_by_state(
    req: &mut Request,
    _res: &mut Response,
) -> Result<String, salvo::Error> {
    let q = req.queries();
    let req_states = match q.get_vec("state") {
        Some(s) => s
            .iter()
            .map(|s| DBState::from_str(s))
            .collect::<Result<Vec<DBState>, _>>()
            .map_err(|e| salvo::Error::Io(std::io::Error::other(e))),
        None => return Err(salvo::Error::Io(std::io::Error::other("No states"))),
    };
    let first_req_state = q.get("state").unwrap();

    assert_eq!(req_states, State::Multiple(vec![DBState::Open, DBState::ClosedCooperative]));
    assert_eq!(first_req_state, DBState::Open);
[...]

PR #1282, opened 2026-01-11
fix: query multi-value extraction for untagged enums

tyreseluo:
## PR Content

- Parse multi-value query/header/form as sequences in deserialize_any.
- Preserve string single-value behavior via deserialize_str/string.
- Add regression test for query untagged enum.

### PR Description

- Problem: query multi-values were parsed via deserialize_any as a single value, so #[serde(untagged)] enums like State::Multiple(Vec<_>) never matched.
- Fix: treat multi-values as sequences in deserialize_any for FlatValue/VecValue, while preserving single-value behavior for String/&str.
- Tests: cargo test -p salvo_core test_de_request_from_query_untagged_mult

## Related Issue/PR

- https://github.com/salvo-rs/salvo/issues/1281

### DIFF
--- a/crates/core/src/serde/flat_value.rs
+++ b/crates/core/src/serde/flat_value.rs
@@ -25,6 +25,21 @@ macro_rules! forward_url_query_parsed_value {
     }
 }
 
+fn looks_like_list(value: &CowValue<'_>) -> bool {
+    let raw = value.0.as_ref().trim();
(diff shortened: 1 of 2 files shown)


tyreseluo:
Hi! @chrislearn 

This PR is work in process. 

Updates to the core crate of Salvo should be handled with caution. I am currently conducting thorough testing to ensure that these updates and fixes do not introduce any issues.

PR #1322, opened 2026-02-22
fix(session): bump RECOMMENDED_KEY_LEN to 64 (512 bits)

AlphaNecron:
`cookie>=0.18` [requires 64 bytes (512 bits)](https://github.com/rwf2/cookie-rs/blob/0.18.0/src/secure/key.rs#L43-L45)
Here's the related excerpt. I wonder why no one has noticed this...
```rust
    /// Creates a new `Key` from a 512-bit cryptographically random string.
    ///
    /// The supplied key must be at least 512-bits (64 bytes).
```

### DIFF
--- a/crates/session/src/lib.rs
+++ b/crates/session/src/lib.rs
@@ -65,7 +65,7 @@
 use std::fmt::{self, Formatter};
 use std::time::Duration;
 
-use cookie::{Cookie, Key, SameSite};
+use cookie::{Cookie, Key, KeyError, SameSite};
 use salvo_core::http::uri::Scheme;
 use salvo_core::{Depot, Error, FlowCtrl, Handler, Request, Response, async_trait};
 use saysion::base64::Engine as _;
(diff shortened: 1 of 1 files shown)


AlphaNecron:
I got some approaches.
- Use `Key::try_from` and handle returned error.
- Remove that `RECOMMENDED_KEY_LEN` and the warning, leave that `Key::from` intact.
- Generate an ephemeral key with `Key::generate` / `Key::try_generate` as stated in the docs. Also warn user to use a persistent key instead.

I prefer 3 but it might defeat the whole purpose of `CookieStore`?

chrislearn:
> I got some approaches.
> 
> * Use `Key::try_from` and handle returned error.
> * Remove that `RECOMMENDED_KEY_LEN` and the warning, leave that `Key::from` intact.
> * Generate an ephemeral key with `Key::generate` / `Key::try_generate` as stated in the docs. Also warn user to use a persistent key instead.
> 
> I prefer 3 but it might defeat the whole purpose of `CookieStore`?

I prefer remove RECOMMENDED_KEY_LEN and add `fn try_new`.

PR #1367, opened 2026-04-01
feat(oapi): use $ref for non-primitive generic type parameters

chrislearn:
## Summary
- Non-primitive type parameters (structs/enums) now use `$ref` instead of being inlined
- Prevents schema duplication in generated OpenAPI output

Ref: utoipa#1184

### DIFF
--- a/crates/oapi-macros/src/component.rs
+++ b/crates/oapi-macros/src/component.rs
@@ -418,6 +418,14 @@ impl ComponentSchema {
                     } else {
                         None
                     };
+                    // Only inline primitive types (String, i32, bool, etc.).
+                    // Non-primitive types (structs, enums) should use $ref
+                    // to avoid schema duplication.
+                    let schema_type = SchemaType {
+                        path: type_path,
+                        nullable,
+                    };
+                    let is_inline = is_inline && schema_type.is_primitive();
                     if is_inline {
                         let default = pop_feature!(features => Feature::Default(_))
                             .map(|feature| feature.try_to_token_stream())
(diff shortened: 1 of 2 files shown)


an-tran511:
`inline(...)` does not fully inline generic schemas

Given:

```rust
#[derive(ToSchema)]
pub struct ResponsePayload<T: ToSchema> {
    data: Vec<T>,
}

#[derive(ToSchema)]
pub struct PrimaryData {
    id: String,
    value: u32,
}
```

and:

```rust
responses(
    (
        status = 200,
        response = inline(ResponsePayload<PrimaryData>)
    )
)
```

I expected the generated `200` response schema to fully inline `PrimaryData` inside `data.items`.

Expected:

```json
{
  "type": "object",
  "properties": {
    "data": {
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "id": {
            "type": "string"
          },
          "value": {
            "type": "integer",
            "format": "int32"
          }
        }
      }
    }
  }
}
```

Actual output still generates `$ref` component schemas:

```json
{
  "components": {
    "schemas": {
      "PrimaryData": { ... },
      "ResponsePayload_PrimaryData": {
        "properties": {
          "data": {
            "items": {
              "$ref": "#/components/schemas/PrimaryData"
            }
          }
        }
      }
    }
  }
}
```

This matters for React type-safe clients because I usually only consume the inner data type on the frontend. Fully inlining the response would avoid generating extra wrapper component schemas and simplify generated client types.

PR #1405, opened 2026-04-16
fix: prevent panic on empty Range header in NamedFile and ReadSeeker

iruizsalinas:
`HttpRange::parse` returns `Ok(vec![])` for range headers like `bytes=` or `bytes=,,,` (valid prefix but no actual range specs). `NamedFile::send` and `ReadSeeker::send` both index `range[0]` without checking, which panics on the empty vec.

Same `!range.is_empty()` guard that `embed.rs` already uses.


chrislearn:
Thanks

PR #1486, opened 2026-05-20
fix(core): detect multipart file parts by filename, not Content-Type

chrislearn:
## Summary

The multipart parser classified a part as a file when the part carried a per-part `Content-Type` header, and otherwise read the body into memory via `field.text().await`. RFC 7578 §4.2 makes the `filename` directive in `Content-Disposition` the indicator of a file part; per-part `Content-Type` is optional metadata.

As a result a binary upload with `Content-Disposition: form-data; name=...; filename=...` but no per-part `Content-Type` was streamed into a `String` instead of written to a temp file. For any deployment that raised \`form_data_max_size\` to accept uploads, this amplified memory pressure and could turn a benign upload into a request-level memory-spike DoS surface.

This change keys the file-vs-text decision on `field.file_name().is_some()`. The `FilePart` continues to capture the per-part `Content-Type` as metadata via `field.headers()`.

## Test plan

- [x] `cargo test -p salvo_core --lib test_form` — 5 passed
- [x] `test_form` extended with a multipart payload that has `filename` but no per-part `Content-Type`; the new assertion verifies it ends up in `files`, not `fields`

## References

- See `_improve.md` (A2) and `_codex_improve.md` for context.

### DIFF
--- a/crates/core/src/http/form.rs
+++ b/crates/core/src/http/form.rs
@@ -90,7 +90,7 @@ impl FormData {
                         ParseError::Multer(e)
                     })? {
                         if let Some(name) = field.name().map(|s| s.to_owned()) {
-                            if field.headers().get(CONTENT_TYPE).is_some() {
+                            if field.file_name().is_some() {
                                 form_data
                                     .files
                                     .insert(name, FilePart::create(&mut field).await?);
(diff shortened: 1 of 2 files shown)


chrislearn:
@copilot resolve the merge conflicts in this pull request

PR #1489, opened 2026-05-20
perf(compression): pre-parse Accept-Encoding entries once

chrislearn:
## Summary

`Compression::negotiate` walked the parsed `Accept-Encoding` list once per algorithm check, calling `.parse::<CompressionAlgo>()` inside `is_accepted`, `is_rejected`, and the client-preference loop. For N server algorithms and M `Accept-Encoding` entries this is O(N × M) string parses on every compressible response.

This change parses each `Accept-Encoding` token once into a `SmallVec<[(CompressionAlgo, u8); 4]>`, capturing the wildcard q-value separately. Downstream lookups become plain `iter().any` / `find_map` calls with no further parsing. SmallVec keeps the typical 1–4-entry list on the stack.

Behavioural change: tokens that fail to parse as a `CompressionAlgo` and are not `*` are now dropped at the parse step rather than being filtered out repeatedly later — semantically identical to the previous code since `.parse::<CompressionAlgo>().ok().as_ref() == Some(algo)` would never match unknown tokens.

## Test plan

- [x] `cargo test -p salvo-compression --lib` — 61 passed, covering wildcards, q=0 rejection, client/server preference, unsupported encoding tokens

## References

- See `_improve.md` (P3) for the audit context.

### DIFF
--- a/crates/compression/Cargo.toml
+++ b/crates/compression/Cargo.toml
@@ -33,6 +33,7 @@ flate2 = { workspace = true, optional = true, features = ["default"] }
 futures-util = { workspace = true }
 indexmap = { workspace = true }
 salvo_core = { workspace = true }
+smallvec = { workspace = true }
 tokio = { workspace = true }
 tokio-util = { workspace = true, features = ["io"] }
 tracing = { workspace = true }
--- a/crates/compression/src/lib.rs
+++ b/crates/compression/src/lib.rs
@@ -376,22 +376,25 @@ impl Compression {
 
         let accept_list = http::parse_accept_encoding(header);
 
-        // Parse each `Accept-Encoding` entry once: drop tokens that map to
-        // neither a wildcard nor a known compression algorithm, so the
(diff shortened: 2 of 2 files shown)


chrislearn:
Good catch on the multi-wildcard regression — pushed [88b6b0f](https://github.com/salvo-rs/salvo/commit/88b6b0f5) to address it.

The fix restores the explicit `accept_list.iter().find(...)` for `wildcard_q` (which yields the first wildcard, i.e. the highest-q one given `parse_accept_encoding` sorts descending), and the parsed-algo collector now just skips `*` entries. Each non-wildcard entry is still parsed at most once, so the perf goal of the PR is preserved.

Added `test_multiple_wildcards_take_highest_q` (locks `Accept-Encoding: *;q=1, *;q=0` ⇒ wildcard fallback still active) — confirmed it would have failed on the previous revision. Full suite stays green (62 passed).

PR #1535, opened 2026-06-18
Tighten proxy forwarding semantics

chrislearn:
## Summary
- reject literal `..` path components while strict path normalization is enabled
- keep opt-out behavior for callers that disable strict path normalization
- document that the default host header getter omits non-default ports and `standard_host_header_getter` preserves them
- document that client IP forwarding overwrites inbound `X-Forwarded-For` instead of trusting/prepending the existing chain

## Validation
- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo check -p salvo-proxy --all-targets --all-features`
- `cargo clippy -p salvo-proxy --all-targets --all-features -- -D warnings`
- `cargo test -p salvo-proxy --all-features`

### DIFF
--- a/crates/proxy/README.md
+++ b/crates/proxy/README.md
@@ -68,10 +68,11 @@ gateway path `/api/users` is forwarded to `http://backend/users`. If the backend
 also expects the `/api` prefix, configure the upstream with that base path:
 `http://backend/api`.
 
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

PR #1547, opened 2026-06-19
Fix session cookie removal attributes

chrislearn:
## Summary
- add `Response::remove_cookie_with` so callers can include removal cookie attributes
- preserve configured session cookie path and domain when destroying a session
- cover the core removal API and session destroy behavior with tests

## Validation
- `cargo fmt --all -- --check` (passes; stable rustfmt reports existing nightly-only option warnings)
- `cargo test -p salvo-session --all-targets`
- `cargo test -p salvo_core --features cookie test_remove_cookie_with_emits_path_and_domain`
- `cargo check --workspace --all-targets --all-features`

### DIFF
--- a/crates/core/src/http/response.rs
+++ b/crates/core/src/http/response.rs
@@ -363,7 +363,19 @@ impl Response {
             self
         }
 
-        /// Removes a cookie from the response.
+        /// Removes a cookie from the response by name.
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

PR #1550, opened 2026-06-20
Fix sliding window rate limiter never sliding

chrislearn:
## Problem

`SlidingGuard::verify` reset `cell_start` to `now` on **every** request (`crates/rate-limiter/src/sliding_guard.rs:69`). The elapsed `delta = now - cell_start` therefore only ever measured the gap since the *previous* request, not the time since the current cell started.

Consequences for steady traffic (gaps smaller than one cell span):
- the `while delta > cell_span` loop never ran, so `head` never advanced;
- old cell counts were never evicted;
- the "sliding" window effectively became a counter that only cleared after a full idle `period`.

The existing tests masked this because `test_sliding_guard_fixed_interval_should_not_reject` manually rewinds `cell_start` before each call, which hides the reset bug.

## Fix

- Anchor cell boundaries in absolute time: advance `cell_start` forward by whole cell spans via `cell_start = now - delta` (the remainder after consuming full spans) instead of resetting to `now`.
- Use `>=` at the `period` and `cell_span` boundaries (off-by-one at exact cell edges).

## Test

Adds `test_sliding_guard_window_slides_with_small_gaps`, which issues requests spaced closer than one cell span but cumulatively exceeding it, and asserts the head advances. This fails on the old code (head stays `0`) and passes now.

All 50 rate-limiter tests pass.

### DIFF
--- a/crates/rate-limiter/src/sliding_guard.rs
+++ b/crates/rate-limiter/src/sliding_guard.rs
@@ -54,20 +54,24 @@ impl RateGuard for SlidingGuard {
             return true;
         }
         let mut delta = now - self.cell_start;
-        if delta > quota.period {
+        if delta >= quota.period {
             self.reset_window(&quota, now);
             return true;
-        } else {
-            while delta > self.cell_span {
-                delta -= self.cell_span;
-                self.head = (self.head + 1) % self.counts.len();
-                self.total = self.total.saturating_sub(self.counts[self.head]);
-                self.counts[self.head] = 0;
-            }
-            self.counts[self.head] += 1;
-            self.total += 1;
-            self.cell_start = now;
         }
+        // Advance the head over every whole cell span that has elapsed since the
+        // current cell started, evicting the counts of the cells that slide out
+        // of the window. Cell boundaries are anchored in absolute time, so
+        // `cell_start` is moved forward by whole cell spans instead of being
+        // reset to `now` on every request (which would freeze the window).
+        while delta >= self.cell_span {
+            delta -= self.cell_span;
+            self.head = (self.head + 1) % self.counts.len();
+            self.total = self.total.saturating_sub(self.counts[self.head]);
+            self.counts[self.head] = 0;
+        }
+        self.counts[self.head] += 1;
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

Thanks for the catch. The cell span is now rounded up by 1ns when `period` isn't evenly divisible by `cells`, so `cells * cell_span >= period` and the ring can no longer wrap and evict a still-valid request before the period elapses. Added `test_sliding_guard_ring_span_covers_period` to lock the invariant.

chrislearn:
Thanks — this is a fair observation, but it describes the **inherent, bounded approximation** of a celled sliding-window *counter*, not a regression introduced here.

Every request inside a bucket shares that bucket's start time, so a burst clustered near the end of a bucket can be dropped up to one `cell_span` early when the bucket is reused. A textbook absolute-time celled counter behaves identically in your example: requests at 0/2.1/2.2/2.3s all fall in absolute cell 0 `[0, 2.5s)`, and the request at 10s maps to absolute cell 4 → the same ring slot, evicting cell 0. The error is bounded by a single bucket's worth of requests and shrinks as `cells` grows.

Eliminating it entirely requires a different algorithm — per-request timestamps (O(limit) memory) or a weighted two-window counter — which changes `SlidingGuard`'s internals and observable behavior. Since the celled model is the public `CelledQuota`/`cells` API and this PR is scoped to the 'window never advances' regression, I've documented the tradeoff on `SlidingGuard` (b1e0bd2) rather than swap the algorithm here. Happy to follow up with a weighted-counter implementation in a dedicated PR if the maintainers want exact accounting.

chrislearn:
@codex review

chrislearn:
@codex review

chrislearn:
@codex review

Good catch on the test flakiness. Reworked `test_sliding_guard_window_slides_with_small_gaps` to use 10 cells over a 2s period (200ms cell span) with six 50ms gaps: the cumulative elapsed time provably advances the head at least once, but cannot wrap back to 0 even under heavy CI sleep overshoot. Now asserts `0 < head < cells`.

PR #1553, opened 2026-06-20
Fix cache swallowing requests and caching non-success responses

chrislearn:
Two correctness bugs in `crates/cache/src/lib.rs`.

### 1. Issuer returning `None` swallowed the request
The `CacheIssuer` docs say returning `None` means "the request will not be cached", but the handler did:

```rust
let Some(key) = self.issuer.issue(req, depot).await else {
    return; // never calls call_next → handler skipped, empty response
};
```

Now it runs the rest of the chain before returning.

### 2. Non-success responses were cached
`cached_response` only excluded streaming/error bodies, so a `500`/`401`/`403` with a normal body was cached and replayed to every client for the TTL. Now only 2xx responses are cached (a missing status code defaults to `200 OK`, so it stays cacheable).

> Note: this restricts caching to 2xx. If you also want to cache specific non-2xx statuses (e.g. `404`, `301`), the `status.is_success()` check is the single place to widen.

### Tests
Adds `test_cache_skips_non_success_status` and `test_cache_issuer_none_runs_handler`. `cargo test -p salvo-cache --lib` → 60 passed.

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -580,11 +580,18 @@ fn cached_response(res: &Response, cache_private: bool) -> Option<CachedEntry> {
     if res.body.is_stream() || res.body.is_error() {
         return None;
     }
-    // Only cache successful responses. A missing status code defaults to `200 OK`
-    // when the response is sent, so treat `None` as cacheable. Caching error or
-    // redirect statuses (e.g. a transient `500`, or an auth-dependent `401`/`403`)
-    // would replay them to every client for the whole TTL.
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

Good catch. `cached_response` now computes the *effective* status the same way `Response::into_hyper` does — an unset status with a `None` body resolves to 404 (and so is not cached), otherwise 200. Added unit tests `test_cached_response_skips_empty_unmatched_404` and `test_cached_response_caches_default_success`.

PR #1554, opened 2026-06-20
Fix tus CORS reflecting arbitrary Origin with credentials

chrislearn:
## Problem

In `crates/tus/src/handlers/mod.rs`, `cors_allow_origin` reflected the request `Origin` when the allowlist was **empty** or contained `*` *and* `allowed_credentials` was enabled:

```rust
if opts.allowed_origins.is_empty() {
    if opts.allowed_credentials {
        return origin.and_then(|o| HeaderValue::from_str(o).ok()); // reflect ANY origin
    }
    ...
}
```

Combined with `Access-Control-Allow-Credentials: true`, this is a credentialed CORS reflection: any website can read upload responses/metadata using the victim's cookies.

## Fix

- An empty/wildcard allowlist now always returns `*` and never reflects an arbitrary origin.
- The `Access-Control-Allow-Credentials` header is omitted whenever the resolved origin is `*` (browsers reject that combination anyway).
- Explicit allowlists are unchanged: a matching origin is still echoed with credentials.

Adds `test_apply_common_headers_never_reflects_with_wildcard_credentials`. `cargo test -p salvo-tus --lib` → 237 passed.

### DIFF
--- a/crates/tus/src/handlers/mod.rs
+++ b/crates/tus/src/handlers/mod.rs
@@ -81,23 +81,27 @@ fn cors_allow_origin(req: &Request, opts: &TusOptions) -> Option<HeaderValue> {
         .get("origin")
         .and_then(|value| value.to_str().ok());
 
-    // An empty or wildcard allowlist never reflects an arbitrary `Origin`.
-    // Reflecting the request origin together with
-    // `Access-Control-Allow-Credentials: true` would let any site read
-    // responses using the victim's cookies, so fall back to `*` (which the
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

Thanks. `cors_allow_origin` now checks the exact allowlist match first, so a request from an explicitly listed origin gets its exact value reflected (with credentials) even when `*` is also present; the `*` fallback only applies to empty/wildcard allowlists and unlisted origins. Added `test_apply_common_headers_listed_origin_wins_over_wildcard`.

PR #1560, opened 2026-06-20
Drop CORS credentials header when Allow-Origin resolves to wildcard

chrislearn:
`ensure_usable_cors_rules` (`crates/cors/src/lib.rs`) only asserts "no `*` with credentials" when credentials are statically `AllowCredentials::Yes` (`is_true()`). A **dynamic** credentials policy (`AllowCredentials::dynamic`/`dynamic_async`) that returns `true` at runtime bypasses every build-time assertion, so combined with `allow_origin(Any)` it can emit:

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
```

— an invalid and unsafe combination (browsers reject it, and it signals a misconfigured guard).

This adds a defense-in-depth check in `CorsHandler::handle`: when the resolved response headers contain credentials `true` **and** origin `*`, the credentials header is dropped and an error is logged. Reflecting a *specific* origin with credentials (the valid pattern) is unaffected.

Adds `test_cors_drops_credentials_with_wildcard_origin`. `cargo test -p salvo-cors --lib` → 29 passed; clippy clean.

### DIFF
--- a/crates/cors/src/lib.rs
+++ b/crates/cors/src/lib.rs
@@ -525,6 +525,26 @@ impl Handler for CorsHandler {
             // This header is applied only to non-preflight requests
             headers.extend(self.cors.expose_headers.to_header(origin, req, depot).await);
         }
+
+        // Defense in depth: `Access-Control-Allow-Origin: *` must never be paired
+        // with `Access-Control-Allow-Credentials: true`. The build-time assertions
+        // in `ensure_usable_cors_rules` only cover statically configured
+        // credentials (`AllowCredentials::Yes`); a dynamic credentials policy can
+        // otherwise produce this invalid, unsafe combination at runtime.
+        if headers
+            .get(header::ACCESS_CONTROL_ALLOW_CREDENTIALS)
+            .is_some_and(|v| v == "true")
+            && headers
+                .get(header::ACCESS_CONTROL_ALLOW_ORIGIN)
+                .is_some_and(|v| v == "*")
+        {
+            tracing::error!(
+                "CORS misconfiguration: `Access-Control-Allow-Credentials: true` cannot be \
+                 combined with `Access-Control-Allow-Origin: *`; dropping the credentials header"
+            );
+            headers.remove(header::ACCESS_CONTROL_ALLOW_CREDENTIALS);
+        }
+
         res.headers_mut().extend(headers);
 
         if self.call_next == CallNext::After {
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

Thanks — extended the runtime guard to mirror all four checks from `ensure_usable_cors_rules`: if credentials resolve to `true` and any of `Access-Control-Allow-Origin`/`-Methods`/`-Headers`/`-Expose-Headers` is `*`, the credentials header is dropped. Added `test_cors_drops_credentials_with_wildcard_methods` (exact origin + wildcard methods on a preflight).

chrislearn:
@codex review

Right again — moved the guard to run after `res.headers_mut().extend(headers)` and it now inspects the final merged response headers (via `get_all`, so a `*` in any position is caught). This covers the default `CallNext::Before` case where a downstream handler set its own CORS headers on `res` before the guard ran. Added `test_cors_drops_credentials_when_handler_sets_wildcard_origin`.

chrislearn:
@codex review

Thanks — moved the guard into `enforce_credentials_safety` and now call it at the very end of `handle`, after both `call_next` points, so it inspects the final response in both `Before` and `After` modes. Added `test_cors_drops_credentials_in_after_mode_with_handler_wildcard` (After mode + handler-set wildcard).

chrislearn:
@codex review

Good point — methods/headers/expose can be comma-separated, so the guard now splits list values on commas and checks each trimmed token for `*` (instead of whole-value equality). Added `test_cors_drops_credentials_with_wildcard_token_in_list` (handler emits `Access-Control-Allow-Headers: *, authorization`).

chrislearn:
@codex review

Agreed — scoped each wildcard check to where the header is meaningful: `Allow-Origin` on every response (security-critical), `Allow-Methods`/`Allow-Headers` only on preflight (OPTIONS), and `Expose-Headers` only on actual responses. This stops a global `Allow-Methods: *` from stripping credentials off valid non-preflight responses. Added `test_cors_keeps_credentials_on_actual_request_with_wildcard_methods` and moved the list-token test to a preflight.

chrislearn:
@codex review

Right — a preflight is OPTIONS **plus** `Access-Control-Request-Method`, so `is_preflight` now requires both; a plain OPTIONS actual request no longer treats `Allow-Methods`/`Allow-Headers` as applicable. Added `test_cors_keeps_credentials_on_options_without_request_method`.

chrislearn:
Thanks — I'm going to leave this one as-is, deliberately.

The mismatch you point out is with the **pre-existing** `if req.method() == Method::OPTIONS` branch (it emits the preflight `Allow-Methods`/`Allow-Headers`/`Max-Age` headers and forces `204`). That branch predates this PR, and changing it to the strict `OPTIONS + Access-Control-Request-Method` predicate would alter the **status-code semantics for every OPTIONS request** (no longer answering bare OPTIONS with `204`), which is a separate behavioral change that real users may depend on — out of scope for a credentials-guard fix and better decided on its own.

The two predicates can also legitimately differ in purpose:
- the emitter branch decides whether to *write* preflight headers — being lenient there is harmless (a browser ignores `Allow-Methods`/`Allow-Headers` on a non-preflight response);
- the guard decides whether a methods/headers wildcard is *security-applicable* — being strict there is the safe direction, since it avoids stripping `Access-Control-Allow-Credentials` from a valid actual response.

So the guard being stricter than the emitter is intentional: it never produces an unsafe `*`+credentials combination on a genuine preflight, and never false-positives on an actual request. Happy to open a separate PR to tighten the core OPTIONS preflight detection (incl. the `204` behavior) if the maintainers want that change.

chrislearn:
@codex review

chrislearn:
@codex review

Fixed — ran `cargo +nightly fmt` on the new tests; `cargo +nightly fmt --all -- --check` is now clean.

PR #1563, opened 2026-06-20
Stop reflecting Access-Control-Request-Headers in tus preflight

chrislearn:
In `crates/tus/src/handlers/options.rs`, when no `allowed_headers` were configured the OPTIONS preflight handler reflected the client's `Access-Control-Request-Headers` value directly into `Access-Control-Allow-Headers`:

```rust
} else if let Some(h) = req.headers().get(H_ACCESS_CONTROL_REQUEST_HEADERS)... {
    headers.insert(H_ACCESS_CONTROL_ALLOW_HEADERS, v); // echoes arbitrary header names
} else { /* DEFAULT_ALLOW_HEADERS */ }
```

Echoing the request header back unchanged mirrors arbitrary attacker-chosen header names, and combined with origin reflection (#1554) widens the CORS attack surface. The handler now always advertises the fixed `DEFAULT_ALLOW_HEADERS` allow-list when no explicit `allowed_headers` are set; configuring `allowed_headers` is unchanged. (Removed the now-unused `H_ACCESS_CONTROL_REQUEST_HEADERS` import.)

`cargo test -p salvo-tus --lib` → 238 passed; clippy clean.

### DIFF
--- a/crates/tus/src/handlers/mod.rs
+++ b/crates/tus/src/handlers/mod.rs
@@ -25,7 +25,7 @@ use crate::{H_TUS_RESUMABLE, TUS_VERSION};
 
 pub(crate) const EXPOSE_HEADERS: &str = "Location, Upload-Offset, Upload-Length, Upload-Metadata, Upload-Expires, Tus-Resumable, Tus-Version, Tus-Extension, Tus-Max-Size";
 pub(crate) const DEFAULT_ALLOW_HEADERS: &str =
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

Good catch. Added `Upload-Defer-Length` (creation-defer-length) and `Upload-Concat` (concatenation) to `DEFAULT_ALLOW_HEADERS` so the fixed allow-list covers every tus request header now that the handler no longer reflects `Access-Control-Request-Headers`.

PR #1568, opened 2026-06-20
Fix serde-util literal unquoting to preserve escaped quotes

chrislearn:
`parse_next_lit_str` (`crates/serde-util/src/lib.rs`) unquoted a string literal with `literal.to_string().replace('"', "")`, which removes **every** `"`, not just the surrounding pair — so `rename = "a\"b"` became `ab` instead of `a"b`. Now parses the token as a `syn::LitStr` and uses `.value()`, which decodes escapes correctly.

> The earlier revision of this PR also switched `snake_case` to an ASCII word boundary, but that diverged from `serde_derive` (which uses Unicode `char::is_uppercase`) and would have made Salvo's generated schema/rename names mismatch the actual serde field names. Per Codex review, that change was reverted; only the literal-unquoting fix remains.

`cargo test -p salvo-serde-util` → 9 passed; clippy clean.

### DIFF
--- a/crates/serde-util/src/case.rs
+++ b/crates/serde-util/src/case.rs
@@ -100,10 +100,7 @@ impl RenameRule {
             SnakeCase => {
                 let mut snake = String::new();
                 for (i, ch) in variant.char_indices() {
-                    // Use ASCII-only detection to match the ASCII-only lowercasing
-                    // below; mixing Unicode `is_uppercase` with `to_ascii_lowercase`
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

Agreed — reverted the snake_case boundary change to keep it aligned with `serde_derive`'s Unicode `is_uppercase` behavior (so generated schema/rename names match serde's actual field names). Only the `parse_next_lit_str` escaped-quote fix remains.

PR #1575, opened 2026-06-20
Back off on accept errors to avoid busy-looping the server

chrislearn:
The accept loop in `crates/core/src/server.rs` logged an accept error and immediately retried:

```rust
Err(e) => { tracing::error!(error = ?e, "accept connection failed"); }
```

A persistent error — most notably `EMFILE`/`ENFILE` when the process is out of file descriptors — makes `accept()` fail instantly on every iteration, spinning the loop at 100% CPU and flooding the logs until descriptors free up. Added a short (10ms) backoff sleep after an accept error so the server degrades gracefully and recovers on its own.

`cargo test -p salvo_core --lib server` → 3 passed; clippy clean.

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -130,6 +130,7 @@ tokio = { workspace = true, features = [
     "macros",
     "net",
     "rt-multi-thread",
+    "time",
 ] }
 tokio-native-tls = { workspace = true, optional = true }
 tokio-openssl = { workspace = true, optional = true }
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

Good catch — `server-handle` is on by default, so the default build uses the `tokio::select!` accept loop. Added the same 10ms backoff to that loop's accept-error branch as well, so both code paths degrade gracefully under EMFILE/ENFILE.

chrislearn:
@codex review

Right — `salvo_core` relied on `tokio/time` being enabled transitively. Added `"time"` explicitly to the tokio features in `crates/core/Cargo.toml`, so the backoff (and the server-handle `tokio::time::Duration` usage) compile under minimal feature sets. Verified `--no-default-features --features http1,server` builds.

PR #1582, opened 2026-06-21
Emit RFC-compliant quoted ETag for embedded files

chrislearn:
`render_embedded_data` (`crates/serve-static/src/embed.rs`) generated the ETag as a **bare** hex hash and checked `If-None-Match` with a whole-string equality:

```rust
let hash = hex::encode(metadata.sha256_hash());
if req.headers().get(IF_NONE_MATCH).map(|e| e.to_str().unwrap_or("000000").eq(&hash)).unwrap_or(false) { ... }
...
if let Ok(etag_val) = hash.parse() { res.headers_mut().insert(ETAG, etag_val); }
```

RFC 7232 entity-tags are **quoted**, and `If-None-Match` is `*` or a comma-separated list of (possibly weak, `W/`-prefixed) tags. A conforming client therefore echoes `"<hash>"` (or `W/"<hash>"`), which never equalled the bare hash — so conditional requests for embedded files almost never returned `304 Not Modified`.

This:
- emits a proper quoted `"<hash>"` ETag;
- parses `If-None-Match` per RFC (`*`, lists, `W/`, quotes), while still accepting the bare legacy form for backward compatibility;
- extracts the match logic into `if_none_match_matches`, covered by a unit test.

`cargo test -p salvo-serve-static --features embed --lib` → 9 passed; clippy clean. The existing `test_serve_embed_files` integration test still passes (it does not assert on the ETag value).

### DIFF
--- a/crates/serve-static/src/embed.rs
+++ b/crates/serve-static/src/embed.rs
@@ -60,17 +60,42 @@ pub fn render_embedded_file(
 
 /// Returns whether an `If-None-Match` header value matches the given hash.
 ///
-/// Per RFC 7232 the header is `*` or a comma-separated list of (possibly weak,
-/// `W/`-prefixed) quoted entity-tags. The bare (unquoted) form is also accepted
-/// for backward compatibility with the previously emitted ETag.
+/// Per RFC 7232 the header is `*` or a list of (possibly weak, `W/`-prefixed)
+/// entity-tags. The opaque value of a quoted tag may itself contain commas, so
+/// the list is parsed quote-aware rather than split on `,`. A bare (unquoted)
+/// tag is also accepted for backward compatibility with the previously emitted
+/// ETag.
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

Good catch — entity-tag opaque values can contain commas, so splitting on `,` was wrong. Replaced it with a quote-aware parser: a quoted tag's opaque value runs to the next `"`, and only commas outside quotes separate list members. Added tests for `"other,<hash>"` (must not match) and a list whose second tag contains a comma (still matches the first).

chrislearn:
@codex review

Good catch — the early `304` return happened before the ETag was inserted, so 304 responses lacked the validator that the corresponding 200/206 carries. Moved the ETag insertion ahead of the `If-None-Match` check so a 304 now echoes the same quoted validator, matching the file-serving path (RFC 7232 §4.1). Extended `test_serve_embed_files` to assert the 304 carries the ETag.

PR #1584, opened 2026-06-21
Clean up endpoint codegen: drop orphan module, fix unused_mut

chrislearn:
## What

Two small, non-behavioral cleanups in the `#[endpoint]` codegen path.

### 1. Delete the orphan `endpoint/server.rs`
`crates/oapi-macros/src/endpoint/server.rs` (462 lines) is not declared by any `mod` statement — `endpoint.rs` only declares `mod attr;`. It was therefore never compiled and is a stale duplicate of the live `src/server.rs` (referenced via `mod server;` in `lib.rs`). Removing it drops dead, drifting code.

### 2. Drop redundant `mut` in the generated modifier block
The generated creator function emits:

```rust
let mut components = &mut components;
let mut operation  = &mut operation;
#(#modifiers)*
```

Both bindings hold `&mut` references that the modifiers use by reborrow and never reassign, so the `mut` keyword is unnecessary. In downstream crates compiled with `#![deny(warnings)]` this surfaced as an `unused_mut` error. Changed to plain `let` bindings.

## Verification
- `cargo build -p salvo-oapi --all-features --tests` — passes; the whole oapi test suite (many `#[endpoint]` usages) still expands and compiles, confirming no modifier reassigns the bindings.


chrislearn:
@codex review

PR #1586, opened 2026-06-21
Don't panic on qualified/generic self types in #[endpoint] impls

chrislearn:
## Problem

The impl-style `#[endpoint]` path derived the name for its generated helper functions with:

```rust
let name = Ident::new(&ty.to_token_stream().to_string(), Span::call_site());
```

`Ident::new` **panics** whenever the string contains characters that are not valid in an identifier. A self type stringifies with spaces and punctuation, so:

- `impl path::Foo { ... }` → `"path :: Foo"` → panic
- `impl Foo<T> { ... }` → `"Foo < T >"` → panic

On top of the panic, even when it didn't crash the reconstructed ident was used in the *type* positions `TypeId::of::<#name>()` / `assign_name::<#name>()`, so a qualified `path::Foo` would have referred to the wrong (unqualified) type.

## Fix

Separate the two distinct uses of the type:

1. **Type-level positions** (`TypeId::of`, `assign_name`) now receive the self type **verbatim** (`#ty`), which is correct for `path::Foo`, `Foo`, and concrete generic instantiations.
2. **Generated helper-fn names** (`__macro_gen_oapi_endpoint_type_id_*` / `_creator_*`) use a new `type_name_suffix(ty)` helper that maps any non-identifier character to `_`, so it never panics.

The `Fn` path is unchanged in behavior — it passes its struct ident for both roles as before.

## Tests
- `type_name_suffix_sanitizes_qualified_and_generic_types` — `Foo`, `path::to::Foo`, `Foo<T>`, `Foo<'a, T>` all yield valid idents; `Foo` is unchanged.
- `endpoint_on_impl_accepts_qualified_self_type` — `impl a::b::Foo { fn handle(&self) {} }` no longer panics and emits `TypeId::of::<a::b::Foo>()` / `assign_name::<a::b::Foo>()`.

## Verification
- `cargo test -p salvo-oapi-macros` — all pass (incl. the two new tests and the existing `endpoint_tests.rs` integration test).
- `cargo build -p salvo-oapi --all-features --tests` — passes.

### DIFF
--- a/crates/oapi-macros/src/endpoint.rs
+++ b/crates/oapi-macros/src/endpoint.rs
@@ -8,11 +8,33 @@ use crate::{Array, DiagResult, InputType, Operation, omit_type_path_lifetimes, p
 mod attr;
 pub(crate) use attr::EndpointAttr;
 
+/// Build a valid identifier usable as a suffix in generated function names from an
+/// arbitrary self type. Any character that is not allowed in an identifier (`:`, `<`,
+/// `>`, `,`, whitespace, ...) is replaced with `_`, so qualified or generic types such
+/// as `path::Foo` or `Foo<T>` no longer panic the way `Ident::new(ty.to_string())` did.
+fn type_name_suffix(ty: &Type) -> Ident {
+    let raw = ty.to_token_stream().to_string();
+    let mut sanitized = String::with_capacity(raw.len());
+    for ch in raw.chars() {
+        if ch.is_ascii_alphanumeric() || ch == '_' {
+            sanitized.push(ch);
+        } else {
+            sanitized.push('_');
+        }
+    }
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Good catch. Generic impls (`impl<T> Foo<T>`) can't be registered through the static inventory at all — the helper fns are non-generic and there's no single concrete type for `TypeId::of`. Rather than emit code referencing the out-of-scope `T`, the impl arm now rejects any impl whose `generics.params` is non-empty with a clear error (`#[endpoint] does not support generic impl blocks; implement it on a concrete type`). Concrete instantiations like `impl Foo<String>` (no params) still work. Added tests for both the rejection and the concrete-instantiation acceptance.

PR #1587, opened 2026-06-21
Elide lifetimes via VisitMut instead of regex string rewriting

chrislearn:
## Problem

`omit_type_path_lifetimes` (duplicated in `salvo_macros` and `salvo-oapi-macros`) anonymized lifetimes by stringifying the type and running a regex:

```rust
let reg = Regex::new(r"'\w+").expect("invalid regex");
let ty_path = ty_path.into_token_stream().to_string();
let ty_path = reg.replace_all(&ty_path, "'_");
syn::parse_str(ty_path.as_ref()).expect("failed to parse type path")
```

Three issues:
1. **Recompiles the regex on every call** (every extractible param of every `#[handler]`/`#[endpoint]`).
2. **Mis-rewrites `'static`** — `'\w+` matches `'static` too, silently turning `Cow<'static, str>` into `Cow<'_, str>`.
3. **Panics** via `expect` if the round-tripped string fails to re-parse.

## Fix

Walk the cloned AST with a `syn::visit_mut::VisitMut` pass and replace each named lifetime with the elided `'_`, leaving `'static` untouched. No stringification, no regex, no panic path.

`regex` was the **only** dependency use in `salvo_macros`, so it is removed from that crate's `Cargo.toml`. `salvo-oapi-macros` keeps `regex` (still used by `schema_type.rs`). Both crates enable syn's `visit-mut` feature.

## Tests
New unit tests in both crates cover:
- `QueryParam<'a, T>` → `QueryParam<'_, T>`
- `Cow<'static, str>` → unchanged
- `Foo<'a, 'static, T>` → `Foo<'_, 'static, T>` (mixed)
- `Vec<String>` → unchanged

## Verification
- `cargo test -p salvo_macros -p salvo-oapi-macros` — all pass.
- `cargo build -p salvo_core --all-features --tests` — passes (`#[handler]` expansion).
- `cargo tree -p salvo_macros -i regex` — regex no longer in the tree.

### DIFF
--- a/crates/macros/Cargo.toml
+++ b/crates/macros/Cargo.toml
@@ -24,9 +24,8 @@ proc-macro = true
 proc-macro-crate = { workspace = true }
 proc-macro2 = { workspace = true }
 quote = { workspace = true }
-regex = { workspace = true }
 salvo-serde-util = { workspace = true }
-syn = { workspace = true, features = ["full", "parsing"] }
+syn = { workspace = true, features = ["full", "parsing", "visit-mut"] }
 
 [lints]
 workspace = true
--- a/crates/macros/src/shared.rs
+++ b/crates/macros/src/shared.rs
@@ -72,7 +72,7 @@ pub(crate) fn parse_input_type(input: &FnArg) -> InputType<'_> {
 /// `'static` intact, so the type can be used in an `Extractible` bound without
 /// carrying caller lifetimes. Operates on the AST via [`VisitMut`] rather than
 /// re-parsing a stringified type, which avoids a per-call regex compile, the
-/// previous `'static` mis-rewrite, and the parse `expect` panic.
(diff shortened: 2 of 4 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Good catch — a type macro carrying the function lifetime (`JsonBody<borrowed!('a)>`) isn't reached by `VisitMut`, since it doesn't parse macro bodies. Added a `visit_macro_mut` that token-rewrites the macro stream, mapping every `'name` except `'static` to `'_` and recursing through groups. This restores the old string-rewrite's macro coverage while still preserving `'static`. Added a test (`JsonBody<borrowed!('a, 'static)>`) in both crates.

chrislearn:
@codex review

PR #1589, opened 2026-06-21
Bind the csrf token by value instead of borrowing a temporary

chrislearn:
## What

`CsrfFinder::find_token` returns an owned `Option<String>`, but `Csrf::handle` matched on a reference to the temporary:

```rust
if let Some(token) = &self.find_token(req).await {
    if !self.cipher.verify(token, &proof) { ... }
```

Binding `token` by value is clearer and idiomatic; `verify` takes `&str`, so the call passes `&token`. No behavior change.

## Verification
- `cargo build/clippy -p salvo-csrf --all-features` — clean.


chrislearn:
@codex review

PR #1591, opened 2026-06-21
Fix RateIssuer::issue doc to match the 400-on-None behavior

chrislearn:
## What

`RateIssuer::issue` was documented as: *"If it returns `None`, the request will not be rate-limited."*

But the handler does the opposite — on `None` it renders `400 Bad Request` (`"invalid identifier"`) and skips the rest of the chain. This is the intended, tested behavior (`test_missing_identifier_renders_consistent_message`).

Reworded the doc to describe the real semantics and to point users at a `Skipper` when they actually want to exempt certain requests from limiting. Docs only.

## Verification
- `cargo doc -p salvo-rate-limiter` with `-D rustdoc::broken_intra_doc_links` — clean.


chrislearn:
@codex review

PR #1592, opened 2026-06-21
Avoid string round-trip in guess_accept_mime

chrislearn:
## What

`guess_accept_mime` recovered an owned `Mime` from `accept.first()` (an `Option<&Mime>`) by stringifying and re-parsing it:

```rust
accept.first().unwrap_or(&dmime).to_string().parse().unwrap_or(dmime)
```

That allocates a `String` and runs the mime parser on every call just to clone a value that's already a valid `Mime`. Replaced with a direct `.cloned()`:

```rust
accept.first().cloned().unwrap_or(dmime)
```

Same result, no allocation or parse. Behavior is identical (a `Mime` re-parsed from its own `to_string()` yields the same `Mime`).

## Verification
- `cargo build -p salvo_core --all-features` — clean.


chrislearn:
@codex review

PR #1594, opened 2026-06-21
Clarify that compression min_length only applies to sized bodies

chrislearn:
## What

`min_length` is only enforced for `ResBody::Once` and `ResBody::Chunks`, whose total length is known before compression. For streaming bodies (`ResBody::Hyper` / `ResBody::Stream`) the length is unknown, so they are compressed unconditionally — meaning a tiny streamed body can still end up *larger* once `Content-Encoding` and framing overhead are added.

This was undocumented and surprising. Added a note to both the `min_length` field and the `min_length(..)` setter explaining the limitation. Docs only — no behavior change (buffering a stream just to measure it would defeat streaming).

## Verification


chrislearn:
@codex review

PR #1595, opened 2026-06-21
Compile the craft attribute regex once via LazyLock

chrislearn:
## What

`take_method_macro` called `Regex::new(REGEX_STR)` every time it processed a `#[craft]` method, recompiling the same attribute-matching regex repeatedly during macro expansion.

Hoisted it into a module-level `LazyLock<Regex>` (the same pattern `salvo-compression` already uses) so it is compiled once per proc-macro process. No behavior change.

## Verification
- `cargo test -p salvo-craft-macros` — passes.


chrislearn:
@codex review

PR #1596, opened 2026-06-21
Document cache key content-negotiation and capacity caveats

chrislearn:
## What

Two undocumented sharp edges in `salvo-cache`:

1. **`RequestIssuer` keys ignore content-negotiation headers.** The key is built from scheme/authority/path/query/method only — not `Accept-Encoding`/`Accept`. With a compression or content-negotiation middleware behind the cache, a client can receive a representation encoded for a *different* request (e.g. a `gzip` body without `Accept-Encoding: gzip`). Documented the risk and the mitigations (custom `CacheIssuer` folding the headers in, ordering the cache ahead of negotiation, or an upstream `Vary` header).

2. **`MokaStore::Builder::max_capacity` counts entries, not bytes.** Cached entries hold full response bodies, so an entry-count bound can still grow unbounded in memory with large responses. Documented this and pointed at a moka `weigher` for byte-based bounds.

Docs only — no behavior change.

## Verification
- `cargo doc -p salvo-cache` with `-D rustdoc::broken_intra_doc_links` — clean.

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -135,6 +135,20 @@ where
 }
 
 /// Identify cacheable requests by their URI.
+///
+/// # Caveats
+///
+/// The generated key is derived only from the request's scheme, authority,
+/// path, query, and (optionally) method. It does **not** include content
+/// negotiation headers such as `Accept-Encoding` or `Accept`. If the cached
+/// responses vary by those headers — for example when a compression middleware
+/// runs behind the cache — a client may receive a representation encoded for a
+/// different request (e.g. a `gzip` body without `Accept-Encoding: gzip`).
+///
+/// When responses are content-negotiated, either keep the cache in front of
+/// (not behind) the negotiating middleware, use a custom [`CacheIssuer`] that
+/// folds the relevant headers into the key, or ensure upstream sets an
+/// appropriate `Vary` header.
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Both fixed:
1. **Ordering reversed.** You're right — hoops run outer-to-inner and the cache captures the response on the way back out, so a cache placed *before* the negotiator stores the already-negotiated variant. Reworded to say the negotiating middleware must run **outside** the cache (added to the router before the cache hoop) so it re-negotiates on every request, including hits.
2. **Weigher now exists.** Rather than point at an unavailable API, I added `MokaStore::Builder::weigher(|k, entry| ...)` so `max_capacity` can bound by total weight (e.g. body bytes). Added `test_weigher_bounds_by_bytes` verifying byte-weighted eviction.

chrislearn:
@codex review

Right — the store doesn't evaluate `Vary` at lookup. Replaced the `Vary` mitigation with a note clarifying that it isn't a fix: under `cache_private(false)` a `Vary` response isn't cached at all, and under `cache_private(true)` it's stored under the same key and replayed regardless of negotiation headers. The two real mitigations (negotiator outside the cache, or a custom `CacheIssuer`) remain.

PR #1597, opened 2026-06-21
Document the disk-DoS surface of FilePart::create

chrislearn:
## What

`FilePart::create` streams an entire multipart field to a temporary file in a loop with **no size limit of its own**. When reached through `Request::form_data`, the body is already bounded by the request's secure-max-size limit (default 64 KiB), so the common path is safe. But calling `create` directly on an otherwise unbounded `Field` lets a client write until disk fills (denial of service).

Added a `# Security` note documenting this and pointing at the existing knobs (`Request::set_secure_max_size`, `set_global_secure_max_size`, the `SecureMaxSize` middleware). Docs only — no behavior change.

## Verification
- `cargo doc -p salvo_core` with `-D rustdoc::broken_intra_doc_links` — the new links resolve (only pre-existing TLS-feature links remain, unrelated to this change).

### DIFF
--- a/crates/core/src/http/form.rs
+++ b/crates/core/src/http/form.rs
@@ -179,6 +179,19 @@ impl FilePart {
 
     /// Create a new temporary FilePart (when created this way, the file will be
     /// deleted once the FilePart object goes out of scope).
+    ///
+    /// # Security
+    ///
+    /// This streams the entire field to a temporary file with **no size limit
+    /// of its own** — it writes until the field ends. Reached through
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Good catch — the bound only holds on the first read. If the body was already cached by an earlier `payload`/`payload_with_max_size` (possibly under a larger limit or none), `form_data` takes the cached-payload path and doesn't re-apply `secure_max_size`. Reworded to scope the guarantee to the first read and to call out the cached-payload / direct-`create` cases explicitly.

PR #1598, opened 2026-06-21
Remove duplicate cfg attribute on the u32 schema-format arm

chrislearn:
## What

In `schema_type.rs`, the `"u32"` match arm under the `non-strict-integers` feature had two identical `#[cfg(feature="non-strict-integers")]` attributes stacked on it:

```rust
#[cfg(feature="non-strict-integers")]
#[cfg(feature="non-strict-integers")]
"u32" => tokens.extend(...UInt32...),
```

Harmless (both must be true, so it's equivalent to one) but a stray copy-paste. Removed the duplicate. No behavior change.

## Verification
- `cargo build -p salvo-oapi-macros` (default) and `--features non-strict-integers` — both build.


chrislearn:
@codex review

PR #1600, opened 2026-06-21
Add Origin checking to WebSocketUpgrade to mitigate CSWSH

chrislearn:
## What

`WebSocketUpgrade::upgrade` performed no `Origin` validation, leaving cookie/session-authenticated WebSocket endpoints open to **Cross-Site WebSocket Hijacking (CSWSH)**: a browser attaches an `Origin` header to cross-site WS handshakes, but unlike `fetch` the same-origin policy does not block the connection, so any site the victim visits could open an authenticated socket using ambient credentials.

Added two opt-in builders:

- `allowed_origins(&["https://app.example.com"])` — case-insensitive exact-match allowlist; a missing or non-matching `Origin` is rejected.
- `check_origin(|origin: Option<&str>| ...)` — custom predicate (receives `None` when the header is absent) for dynamic policies or to explicitly permit a missing `Origin`.

When a policy is configured and the handshake fails it, `upgrade` returns `403 Forbidden` **before** switching protocols. Default behavior is unchanged (no check unless configured). The method docs carry the CSWSH rationale.

## Tests
- `origin_check_allowlist_and_predicate` — unit-tests the matcher (case-insensitivity, missing-Origin rejection, predicate).
- `test_websocket_rejects_disallowed_origin` — a handler with an allowlist rejects a disallowed `Origin` with `403`.

## Verification
- `cargo build -p salvo_extra --features websocket` — compiles.
- `cargo clippy -p salvo_extra --features websocket --lib` + `cargo +nightly fmt --all -- --check` — clean.
- Note: the `salvo_extra` **test** binary can't link locally on Windows because the dev-dependency pulls in `salvo_core`'s `rustls`/`aws-lc-rs` stack (pre-existing, unrelated to this change); the new tests run in CI on Linux.

### DIFF
--- a/crates/extra/src/websocket.rs
+++ b/crates/extra/src/websocket.rs
@@ -92,7 +92,7 @@ use futures_util::sink::{Sink, SinkExt};
 use futures_util::stream::{Stream, StreamExt};
 use futures_util::{FutureExt, TryFutureExt, future};
 use hyper::upgrade::OnUpgrade;
-use salvo_core::http::header::{SEC_WEBSOCKET_PROTOCOL, SEC_WEBSOCKET_VERSION, UPGRADE};
+use salvo_core::http::header::{ORIGIN, SEC_WEBSOCKET_PROTOCOL, SEC_WEBSOCKET_VERSION, UPGRADE};
 use salvo_core::http::headers::{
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

PR #1602, opened 2026-06-21
Return Cow from RenameRule::apply_to_field to skip no-op allocations

chrislearn:
## What

`RenameRule::apply_to_field` always returned an owned `String`. For the `snake_case`/`lowercase` rules the field name is unchanged, yet the old code still did `field.to_owned()`. The request deserializer calls this **per renamed field** when `rename_all` is configured, so those no-op rules allocated needlessly on the hot path.

Changed the return type to `Cow<'_, str>`: `snake_case`/`lowercase` now return `Cow::Borrowed(field)` (zero allocation); every other rule returns `Cow::Owned` exactly as before.

## Breaking change (0.94)

`extract::RenameRule` is publicly re-exported, so the return-type change is breaking for external callers. Migration is mechanical:
- need an owned `String`? → `.into_owned()`
- `&*x`, `x.to_string()`, deref-coercion to `&str` → unchanged

The only in-tree callers are core's request deserializer, updated here. (The macros use a separate `serde-util` `RenameRule`, unaffected.)

## Verification
- `cargo test -p salvo_core --all-features --lib extract::case` and `serde::request` — pass.

### DIFF
--- a/crates/core/src/extract/case.rs
+++ b/crates/core/src/extract/case.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::str::FromStr;
 
 use self::RenameRule::{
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

PR #1603, opened 2026-06-21
Generate craft handle types via split_for_impl for generic impls

chrislearn:
## Problem

The `#[craft]` macro generates a private `handle` struct parameterized by the **impl's** generics, then references it in impl headers. It derived the type arguments for those references from `self_ty`:

```rust
// angle_bracketed taken from self_ty
impl #impl_generics Deref for handle #angle_bracketed #where_clause { ... }
impl #impl_generics handle #angle_bracketed #where_clause { #method }
```

`self_ty`'s arguments are not the same as the `handle` struct's parameters whenever they differ from the impl's generic params. For `impl<T> Pair<T, T>`, `self_ty = Pair<T, T>` so it emitted `handle<T, T>` — but `handle` has a single parameter `<T>`, so the generated code failed to compile.

## Fix

Use `Generics::split_for_impl()`:
- `impl_g` for the `impl<T: Bound>` headers (and the struct definition),
- `ty_g` for type references (`handle<T>`),
- `where_c` for the where-clause.

This drops the manual `self_ty`-argument extraction and the redundant `make_where_clause`. Non-generic and simple-generic impls are unaffected; the broken cross-arg cases now compile.

## Tests
- New `crafted_handler_supports_generic_impl_with_repeated_type_args` in `craft_smoke.rs` exercises `impl<T> Pair<T, T>`, which fails to compile under the old codegen and passes now.
- Existing craft smoke / oapi-doc tests still pass.

## Verification
- `cargo test -p salvo-craft` — all pass.

### DIFF
--- a/crates/craft-macros/src/craft.rs
+++ b/crates/craft-macros/src/craft.rs
@@ -5,8 +5,7 @@ use quote::{ToTokens, quote};
 use regex::Regex;
 use syn::parse::Parser;
 use syn::{
-    AngleBracketedGenericArguments, Attribute, FnArg, Generics, Ident, ImplItem, ImplItemFn, Item,
-    PathArguments, Token, Type, TypePath, parse_quote,
+    Attribute, FnArg, Generics, Ident, ImplItem, ImplItemFn, Item, Token, Type, parse_quote,
 };
 
 use crate::utils::salvo_crate;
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

PR #1604, opened 2026-06-21
Clarify that dropping certon's maintenance handle detaches, not cancels

chrislearn:
## What

`AcmeListener::try_bind` discards the handle from `certon::start_maintenance`:

```rust
let _maintenance_handle = certon::start_maintenance(&certon_config);
```

This looks like a lifetime bug ("the background renewal task gets cancelled when the handle drops"), but `start_maintenance` returns a `tokio::task::JoinHandle`, and certon's own docs state that **dropping the handle detaches the task rather than cancelling it** (`CertCache::stop` is the way to request shutdown). So the renewal/OCSP loop keeps running for the process lifetime.

Added a comment at the call site documenting this so the intentional drop isn't repeatedly re-flagged as a bug. No behavior change.

## Verification
- `cargo build -p salvo-acme` — compiles.
- `cargo +nightly fmt --all -- --check` + `typos` — clean.

### DIFF
--- a/crates/acme/src/listener.rs
+++ b/crates/acme/src/listener.rs
@@ -402,7 +402,10 @@ where
         let tls_acceptor = TlsAcceptor::from(server_config.clone());
         let inner = inner.try_bind().await?;
 
-        // Start certon's background maintenance for renewal + OCSP.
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

PR #1605, opened 2026-06-21
Default Proxy to the standards-compliant host header getter

chrislearn:
## What

`Proxy::new` and `Proxy::with_client_ip_forwarding` defaulted to `default_host_header_getter`, which forwards only the bare upstream host and **drops a non-default port** — `http://upstream:8080` becomes `Host: upstream`. Upstreams that route/virtual-host by `Host:port` get the wrong `Host`.

Switched the default to `standard_host_header_getter`, which includes the port when it isn't the scheme default, per RFC 7230 / RFC 9110. The previous behavior stays available via `default_host_header_getter` configured through `Proxy::host_header_getter`.

## Behavior change (0.94)

For upstreams with a non-default port, the forwarded `Host` now includes the port. Most servers accept this; deployments that relied on the bare-host behavior can opt back in with `default_host_header_getter`.

## Tests
- New `test_proxy_default_host_header_getter_includes_port` asserts `Proxy::new(...)`'s default getter keeps `host.tld:8080`.
- Existing `test_host_header_handling` (both getters) and all 31 proxy tests pass.

## Verification
- `cargo test -p salvo-proxy --all-features` — pass.

### DIFF
--- a/crates/proxy/README.md
+++ b/crates/proxy/README.md
@@ -93,9 +93,10 @@ only prevents common proxy/backend path interpretation mismatches.
 
 ### Forwarded headers
 
-The default host header getter forwards only the upstream host name. Configure
-`standard_host_header_getter` when the upstream `Host` header should include a
-non-default port, for example `example.com:8080`.
+The default host header getter (`standard_host_header_getter`) forwards the
+upstream host and includes a non-default port, for example `example.com:8080`
+(per RFC 7230 / RFC 9110). Configure `default_host_header_getter` if you want to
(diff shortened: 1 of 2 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Good catch — updated `crates/proxy/README.md` 'Forwarded headers' section to state the default is now `standard_host_header_getter` (includes the port) and that `default_host_header_getter` is the opt-in for bare-host forwarding, matching the code change.

PR #1606, opened 2026-06-21
Bound the protocol-detection read to mitigate idle-connection leaks

chrislearn:
## Problem

`serve_connection` detects HTTP/1 vs HTTP/2 by reading the first bytes (`read_version`). On the `http1 + http2` path:

```rust
if let Some(fusewire) = &fusewire {
    tokio::select! { result = read_version(socket) => ..., _ = fusewire.fused() => return }
} else {
    read_version(socket).await?   // <-- no timeout
}
```

With a fusewire the read is bounded by the fuse, but **fusewire is `None` by default**, so the bare `read_version(socket).await` has no timeout. A client that opens a TCP connection and never sends a byte keeps that detection read pending forever — a Slowloris-style way to leak server tasks / exhaust connections.

## Fix

Wrap the no-fusewire detection read in a generous default timeout (`30s`); on elapse the connection is closed. This:
- bounds only the **initial** protocol-detection read (real clients send the request line / HTTP/2 preface immediately), not established or keep-alive connections;
- leaves the fusewire path unchanged (it already times out via its `select!`);
- points users at `Server::fuse_factory` for finer-grained handshake/idle timeouts.

## Behavior change

A connection that opens and sends nothing for 30s is now closed rather than held indefinitely.

## Verification
- Builds with `http1+http2`, and with `http1`-only (the const is `cfg`-gated to the both-features path).
- `cargo test -p salvo_core --features "server http1 http2 test" --lib conn` — passes (clients in tests send promptly, well within 30s).
- fmt clean. (No fast unit test for the 30s path; the change is a localized `tokio::time::timeout` wrapper.)

### DIFF
--- a/crates/core/src/conn/proto.rs
+++ b/crates/core/src/conn/proto.rs
@@ -39,6 +39,11 @@ const H2_PREFACE: &[u8] = b"PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
 /// immediately — and only bounds this one initial read, not established connections.
 /// Configure a [`fuse_factory`](crate::Server::fuse_factory) for finer-grained
 /// handshake/idle timeouts.
+///
+/// This relies on the Tokio **time driver**, so the server must run on a runtime
+/// built with timers enabled (`#[tokio::main]` and `Runtime::new()` enable them;
+/// a hand-built runtime needs `enable_time()` / `enable_all()`). That matches the
+/// rest of the server (graceful-shutdown and fuse timeouts already use timers).
 #[cfg(all(feature = "http1", feature = "http2"))]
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Fair point. There's no stable API to detect whether the time driver is enabled, so I can't degrade gracefully at runtime. I documented the requirement on the const: the server needs a runtime with timers (`#[tokio::main]`, `Runtime::new()`, or `enable_time()`/`enable_all()`). This matches the rest of the server, which already relies on the time driver for graceful-shutdown and fuse timeouts, and matches the norm for Tokio-based servers (axum/hyper/actix similarly assume timers). If you'd prefer this default left to explicit opt-in instead, I can gate it behind a builder knob — let me know.

PR #1607, opened 2026-06-21
Add Server::max_connections to bound concurrent connections

chrislearn:
## What

Adds an opt-in `Server::max_connections(n)` that caps the number of concurrent connections, addressing the connection-exhaustion half of the Slowloris/DoS concern (the protocol-detection read timeout in #1606 handled the idle-read half).

Implementation: a `tokio::sync::Semaphore` with `n` permits. The accept loop acquires a permit **before** accepting and the spawned connection task holds it for the connection's lifetime, releasing it on close. Once `n` connections are active, the loop stops accepting — applying backpressure to the OS listen backlog — until one closes. Default is unlimited (no semaphore, zero overhead).

### Shutdown safety

In the `server-handle` build the permit acquisition is `select!`'d against the graceful and force stop tokens (`biased`), so a saturated server still breaks out of the accept loop on shutdown instead of blocking forever waiting for a permit. The non-`server-handle` build has no stop channel, so a plain blocking acquire is used.

## Tests
- `test_server_max_connections_still_stops` — a `max_connections(1)` server stops within 1s for both graceful and forceful stop, verifying the permit/stop-token interplay can't deadlock the accept loop.
- Existing server tests still pass; builds with and without `server-handle`.

## Verification
- `cargo build -p salvo_core` with `server-handle` and without — both compile.
- `cargo test -p salvo_core ... --lib server` — pass; clippy + fmt clean.

### DIFF
--- a/crates/core/src/server.rs
+++ b/crates/core/src/server.rs
@@ -25,6 +25,8 @@ use tokio::{
 #[cfg(feature = "server-handle")]
 use tokio_util::sync::CancellationToken;
 
+use tokio::sync::Semaphore;
+
 use crate::Service;
 #[cfg(feature = "quinn")]
 use crate::conn::quinn;
@@ -119,6 +121,8 @@ pub struct Server<A> {
     acceptor: A,
     builder: HttpBuilder,
     fuse_factory: Option<ArcFuseFactory>,
+    /// Maximum number of concurrent connections; `None` means unlimited.
+    max_connections: Option<usize>,
     #[cfg(feature = "server-handle")]
     tx_cmd: UnboundedSender<ServerCommand>,
     #[cfg(feature = "server-handle")]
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Both fixed:
1. **(P1) Saturated shutdown deadlock** — you're right, the permit-acquire `select!` only raced the stop *tokens*, which are cancelled in the accept select's `rx_cmd` arm that's unreachable while parked in `acquire_owned`. Now the permit-acquire select polls `rx_cmd.recv()` directly too (biased), via a shared `handle_stop_command!` macro, so a saturated server breaks the loop on stop. Added `test_server_max_connections_stops_while_saturated`: it holds the single permit with a long-lived connection, then force-stops, and asserts shutdown completes (this would hang under the old code).
2. **(P2) Test cfg gate** — gated the test on `server-handle`.

PR #1608, opened 2026-06-21
salvo-core: Improve preformance for `FlowCtrl`

cnlancehu:
### Changes made in this PR                                                                                                                                                                                                                                                                                                                                                                                                                                       
- Remove redundant `DEFAULT_STATUS_OK_HANDLER` — It sets status to 200 OK after the goal handler, but the post-chain check (service.rs:285-288) already does the same. Saves one Arc clone and one Vec slot per request.                                                                                                                                                                                     
- Avoid per-request Arc clones in `FlowCtrl::call_next` — Replace `Vec<Arc<dyn Handler>>` with `Vec<Option<Arc<dyn Handler>>>` so handlers can be `.take()` out of the slot instead of `.cloned()`. This eliminates atomic refcount operations for every handler fetch in call_next, reducing total atomic ops.

### DIFF
--- a/crates/core/src/conn/proto.rs
+++ b/crates/core/src/conn/proto.rs
@@ -112,12 +112,13 @@ impl HttpBuilder {
         } else {
             // No fusewire: still bound the protocol-detection read so a connection
             // that never sends bytes can't leak a task indefinitely.
-            match tokio::time::timeout(PROTOCOL_DETECT_READ_TIMEOUT, read_version(socket)).await {
-                Ok(result) => result?,
-                Err(_) => {
-                    tracing::info!("closing connection: protocol-detection read timed out");
-                    return Ok(());
-                }
+            if let Ok(result) =
+                tokio::time::timeout(PROTOCOL_DETECT_READ_TIMEOUT, read_version(socket)).await
+            {
+                result?
+            } else {
+                tracing::info!("closing connection: protocol-detection read timed out");
+                return Ok(());
             }
         };
         #[cfg(all(not(feature = "http1"), not(feature = "http2")))]
--- a/crates/core/src/handler.rs
+++ b/crates/core/src/handler.rs
@@ -395,14 +395,13 @@ impl Handler for HoopedHandler {
     ) {
         let inner: Arc<dyn Handler> = self.inner.clone();
         let right = ctrl.handlers.split_off(ctrl.cursor);
-        ctrl.handlers.append(
-            &mut self
-                .hoops
+        ctrl.handlers.extend(
+            self.hoops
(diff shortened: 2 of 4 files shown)


chrislearn:
Thanks for the perf work! The two changes here are quite different in nature, so I'd suggest splitting this PR.

✅ FlowCtrl — Option::take instead of Arc::clone

This part looks good. Replacing Vec<Arc<dyn Handler>> with Vec<Option<Arc<dyn Handler>>> and .take()-ing handlers avoids the atomic refcount bump on every fetch in call_next, and it also removes the possibility of a handler being dispatched twice that the old .cloned() path technically allowed. Could you add a small test covering re-entrant call_next (around-style middleware calling call_next) so we lock in the behavior?

❌ Removing DEFAULT_STATUS_OK_HANDLER — this is a regression

This one isn't redundant. The post-chain check at service.rs:285-288 is not equivalent, because of middleware unwinding order.

The chain is [service hoops…, router hoops…, DefaultStatusOK, goal], and DefaultStatusOK calls call_next internally, so it wraps only the goal:

goal runs (writes body, no status)
  → DefaultStatusOK sets status = 200 OK        ← here
    → hoops' post-call_next code unwinds, sees Some(200)  ✓
After removing it:

goal runs (no status)
  → hoops' post-call_next code unwinds, sees None         ✗
    → the 285-288 check sets 200 only afterwards (too late for hoops)
So any middleware that reads res.status_code after call_next (access logging, metrics, compression decisions, etc.) will observe None instead of the implicit 200 whenever the goal handler relied on the default.

This handler was added on purpose in [#925](https://github.com/salvo-rs/salvo/issues/925) ("service hoops can not get status code correctly") — the comment We hope all hoops in service can get the correct status code. is exactly about this. Removing it reintroduces that bug, and there's currently no test guarding it.

Suggestion: keep the FlowCtrl optimization, drop the DefaultStatusOK removal. Happy to add a regression test for the #925 scenario if useful.

cnlancehu:
Thanks a lot for the detailed explanation.

I've reverted the commit that removed DEFAULT_STATUS_OK_HANDLER—that was a bit too hasty on my part. Sorry about that.

The PR now contains the following changes:

- Replace Vec<Arc<dyn Handler>> with Vec<Option<Arc<dyn Handler>>> and use Option::take() in FlowCtrl
- Add tests covering the updated FlowCtrl behavior
- Add a regression test for #925 to ensure service-level middleware can observe the implicit 200 OK status after call_next()
- Fix a Clippy warning in HttpBuilder::serve_connection()

Please let me know if you'd like any further adjustments.

PR #1615, opened 2026-06-25
fix(security): harden proxy IP trust, JWKS cache bounds, ACME and redirect fail-closed

chrislearn:
## Background

Part of a project-wide audit, this PR groups confirmed "security hardening" fixes. Each was verified by reading the source; the affected crates build and test cleanly.

### 1. rate-limiter: `TrustedProxyIssuer` could be bypassed with a spoofed IP (high severity)
`extract_forwarded_ip` took the **leftmost** `X-Forwarded-For` entry as the client IP. `X-Forwarded-For` is append-only, so the leftmost value is written by the client and fully attacker-controlled: an attacker could rotate forged IPs to bypass per-IP rate limiting — exactly what `TrustedProxyIssuer` is supposed to prevent.

It now scans the header **right-to-left**, skipping configured trusted proxies, and returns the first non-trusted address (the real client). Two regression tests are added: a spoofed leftmost value, and a chain of trusted proxies. The legacy `ForwardedHeaderIssuer` (documented as unconditionally trusting) is unchanged.

### 2. jwt-auth: JWKS/OIDC `Cache-Control` directives were unbounded
`max-age`/`stale-*` parsed from the IdP had no bounds: `max-age=0` forced revalidation on every request (amplifying load back to the IdP), and a huge value could overflow the downstream `fetched + max_age` arithmetic (debug panic / release wrap). `max-age` is now clamped to `[1s, 30d]`; `stale-*` are capped at 30d.

### 3. acme: directory URL had no scheme check
A misconfigured `http://` directory leaves the entire ACME exchange open to MITM tampering. A non-HTTPS directory URL now emits a `warn` (http is kept for test environments such as pebble).

### 4. force_https: fail-open on redirect build failure
When the redirect URI failed to build, the old code silently fell through and served the plaintext request, defeating the middleware. It now fails closed (returns 400 and `skip_rest`).

### 5. tus: `is_safe_upload_id` had no length cap
Adds a 256-byte cap to bound the resulting file-path length and limit abuse with attacker-chosen IDs.

## Testing
- `cargo test -p salvo-rate-limiter -p salvo-jwt-auth -p salvo_extra -p salvo-tus -p salvo-acme` — all pass (including the two new rate-limiter regression tests).

### DIFF
--- a/crates/acme/src/config.rs
+++ b/crates/acme/src/config.rs
@@ -107,9 +107,17 @@ impl AcmeConfigBuilder {
     #[inline]
     #[must_use]
     pub fn directory(self, name: impl Into<String>, url: impl Into<String>) -> Self {
+        let url = url.into();
+        if !url.starts_with("https://") {
+            tracing::warn!(
+                directory_url = %url,
+                "ACME directory URL is not HTTPS; the ACME exchange is not protected against \
+                 man-in-the-middle tampering. Use an https:// directory in production."
+            );
+        }
         Self {
             directory_name: name.into(),
-            directory_url: url.into(),
+            directory_url: url,
             ..self
         }
     }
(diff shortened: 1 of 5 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

PR #1617, opened 2026-06-25
fix: repair silently-broken request id, extract flatten and cache headers

chrislearn:
## Background

Part of a project-wide audit, this PR groups confirmed "silent failure" bugs — features that compile but do not actually work. Each fix was verified by reading the source and building/testing the affected crate.

### 1. `extra/request_id`: `depot.request_id()` always returned `None`
The middleware stored the id in the depot as a `HeaderValue`, but `RequestIdDepotExt::request_id()` looks it up as a `String`. The type mismatch made the public accessor always return `None`. The existing test masked this by reading the value back as `HeaderValue`.

The id is now stored as a `String`, and the test exercises the real `request_id()` accessor.

### 2. `macros/extract`: `flatten` could be silently reset
With multiple `#[salvo(extract)]` attributes on a field, an unconditional `flatten = info.flatten` let a later attribute without `flatten` reset an earlier `flatten = true` back to `None`. The override is now gated on the attribute actually specifying `flatten` (matching how `rename` is handled).

### 3. `cache`: cached response replaced the entire header map
`respond_from_cache` did `*res.headers_mut() = headers`, discarding headers set by outer middleware (request id, CORS, security headers). It now merges: same-named cached headers fully replace existing ones (keeping multi-valued headers like `Set-Cookie` intact), while unrelated headers are preserved.

### 4. `cache`: `Cache-Control: no-cache` is now treated as non-cacheable
`no-cache` requires revalidation before reuse. This middleware does not revalidate, so serving such a response from cache is incorrect; it is now excluded from caching alongside `no-store`/`private`.

## Testing
- `cargo build -p salvo_extra -p salvo-cache -p salvo_macros` — clean.
- `cargo test -p salvo-cache --all-features` — pass.
- `cargo test -p salvo_extra ... --all-features` — pass, including `request_id` tests now driven through the public accessor.

### DIFF
--- a/crates/cache/src/lib.rs
+++ b/crates/cache/src/lib.rs
@@ -558,7 +558,9 @@ where
     /// Allow caching requests or responses that contain private-user cache signals.
     ///
     /// By default, requests with `Authorization` or `Cookie`, and responses with `Set-Cookie`,
-    /// `Vary`, or `Cache-Control: private/no-store`, are not cached.
+    /// `Vary`, or `Cache-Control: private`, are not cached. Enabling this lifts those
+    /// privacy restrictions. `Cache-Control: no-store`/`no-cache` responses are never
+    /// cached regardless of this setting.
     #[inline]
     #[must_use]
     pub fn cache_private(mut self, cache_private: bool) -> Self {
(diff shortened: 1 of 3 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

chrislearn:
@codex review

chrislearn:
@codex review

PR #1619, opened 2026-06-25
perf(core): single map lookup for depot value extraction

chrislearn:
## Background

Part of a project-wide audit, this PR addresses a hot-path inefficiency in depot-based extraction.

### 1. `get_depot_value`: 19 map lookups → 1
Extracting a field from the depot (`#[salvo(extract(source(from = "depot")))]`) previously iterated a table of 19 closures, each calling `depot.get::<T>(key)` — i.e. up to 19 `HashMap<String, _>` lookups (each hashing the key) per field.

A new `pub(crate) Depot::get_any` returns the type-erased value, so `get_depot_value` now does **one** lookup and then probes the candidate types with cheap `downcast_ref`s (a `TypeId` comparison). A boxed value has exactly one concrete type, so probe order does not affect the result — behaviour is unchanged.

### 2. `Depot::get_mut`: redundant downcast removed
`get_mut` called `downcast_mut::<V>()` twice (once to test, once to return). It now checks `is::<V>()` then downcasts once, consistent with `get_typed_mut`.

## Testing
- `cargo test -p salvo_core --lib` — 205 tests pass, no regression (depot + serde request deserializer covered).

### DIFF
--- a/crates/core/src/depot.rs
+++ b/crates/core/src/depot.rs
@@ -221,6 +221,15 @@ impl Depot {
         }
     }
 
+    /// Borrow the type-erased value stored under `key` via [`Depot::insert`], if any.
+    ///
+    /// Lets a caller that probes many concrete types do a single map lookup and
+    /// then `downcast_ref` repeatedly, instead of one lookup per candidate type.
+    #[inline]
+    pub(crate) fn get_any(&self, key: &str) -> Option<&(dyn Any + Send + Sync)> {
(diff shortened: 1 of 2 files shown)


chrislearn:
Good catch — ran `cargo +nightly fmt` on the touched files (`depot.rs` and `serde/request.rs`); `cargo +nightly fmt --all -- --check` is now clean.

chrislearn:
@codex review

PR #1625, opened 2026-06-29
fix(core): FlatParser offset, CombWisp wild capture, multipart predicate

chrislearn:
Extraction / serde / HTTP correctness fixes in core from the second-round audit.

## Changes

- **serde `FlatParser` offset** — `FlatParser::new` hardcoded `start = 1`, assuming byte 0 is `[`. But `looks_like_list` only checks the *trimmed* value, so a list value with leading whitespace (`?x=  [1,2,3]`) scanned from the wrong offset, and a multi-byte char before `[` made `get(1..)` land mid-codepoint and silently yield an empty sequence. Locate the actual `[`. Adds a leading-whitespace test.
- **routing `CombWisp` wild capture** — the wild remaining path was snapshotted via `all_rest()` *before* `forward()` consumed the comb part, then a `strip_prefix` hack tried to remove the captured values (which were not a prefix). `{**rest}` (and a constrained `{**rest|\d+}`) captured/validated the still-consumed prefix (e.g. `first123world.ext` instead of `.ext`). Read `all_rest()` *after* `forward()`. Tests now assert the captured values, not just that `detect()` is true.
- **multipart content-type predicate** — `form_data*` accepted any `multipart/*` while `parse_form`/`parse_body*` only accepted `multipart/form-data`, so two equivalent APIs diverged on exotic `multipart/*` subtypes. Unify behind `is_form_content_type`.

(An earlier `write_body` change was reverted after review — the catcher deliberately relies on `ResBody::Error` being overwritten by `write_body` to render error pages.)

## Testing

`cargo test -p salvo_core --lib` (207) passes; path tests pass under `matched-path`; new/strengthened FlatParser and CombWisp tests; `cargo fmt`.

### DIFF
--- a/crates/core/src/http/request.rs
+++ b/crates/core/src/http/request.rs
@@ -1251,7 +1251,7 @@ impl Request {
     #[inline]
     pub async fn form_data_max_size(&mut self, max_size: usize) -> ParseResult<&FormData> {
         if let Some(ctype) = self.content_type() {
-            if ctype.subtype() == mime::WWW_FORM_URLENCODED || ctype.type_() == mime::MULTIPART {
+            if is_form_content_type(&ctype) {
                 let body = self.take_body();
                 if body.is_none() {
                     let bytes = self.payload_with_max_size(max_size).await?.to_owned();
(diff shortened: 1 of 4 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

PR #1627, opened 2026-06-29
fix(oapi): Number Eq soundness, macro error span, drop unused regex, doc fixes

chrislearn:
OpenAPI and proc-macro fixes from the second-round audit.

## Changes

- **`schema::Number` unsound `Eq`** — `Float` compared with `f64 == f64`, so `Number::Float(NaN)` was unequal to itself, violating the reflexivity `Eq` requires. Compare floats as "numerically equal OR same bit pattern" so `NaN == NaN` holds and `-0.0 == 0.0` still holds. Adds a test. (Cross-variant distinctness `Int(1) != UInt(1)` is kept — there is an existing test for it.)
- **oapi-macros error span** — the invalid-status-range error used `value.span()` where `value` is a `String`; `String::span()` resolves to `Span::call_site()`, mislocating the diagnostic to the macro call site. Use `lit_str.span()`.
- **oapi-macros message** — the `#[endpoint]` fallback error read `#[handler] must added ...`; corrected to `#[endpoint] must be added ...`.
- **drop unused `regex` dependency** — the lifetime-omission pass was rewritten to a `syn` visitor; no `regex` use remains in oapi-macros.
- **docs** — `Schema::AllOf` said "AnyOf"; `Paths::append` claimed whole-value overwrite while `insert` merges field by field — both corrected.

(An earlier change aligning the form body media type to `multipart/form-data` was reverted after review — it collided with `FormFile`/`FormFiles::register`, clobbering the form-field schema; kept as `multipart/*`. Proper `multipart/form-data` needs schema merging — left as follow-up. The form `to_schema` is still built once and reused.)

## Testing

`cargo test -p salvo-oapi --lib` and `cargo test -p salvo-oapi-macros` pass (new Number NaN-reflexivity test). Note: salvo-oapi has 2 pre-existing failing tests on main (`test_openapi_schema_work_with_generics`, `test_simple_document_with_security`, `uint16` vs `int32`) unrelated to this PR.

### DIFF
--- a/crates/oapi-macros/Cargo.toml
+++ b/crates/oapi-macros/Cargo.toml
@@ -39,7 +39,6 @@ proc-macro2 = { workspace = true }
 quote = { workspace = true }
 proc-macro-crate = { workspace = true }
 proc-macro2-diagnostics = { workspace = true }
-regex = { workspace = true }
 salvo-serde-util = { workspace = true }
 syn = { workspace = true, features = ["full", "extra-traits", "visit-mut"] }
 salvo_core = { workspace = true }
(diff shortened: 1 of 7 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

PR #1628, opened 2026-06-29
fix(jwt-auth, flash): OIDC h2 negotiation, clock-skew panic, panic-free flash accessors

chrislearn:
jwt-auth / flash fixes from the second-round audit.

## Changes

- **OIDC client HTTP/2** — the default discovery/JWKS client only `enable_http1()`; some IdP endpoints (often CDN-fronted) prefer/require h2 via ALPN, so an http1-only connector failed the handshake. Use `enable_http1().enable_http2()` and add the `http2` feature to hyper-rustls so the method is available.
- **`current_time()` clock-skew panic** — called on the token-validation hot path; a clock set before the epoch (NTP step / VM time travel) made `expect("Time Went Backwards")` panic. Fall back to `0` with `unwrap_or_default()`.
- **flash panic-free accessors** — add `try_outgoing_flash`/`try_outgoing_flash_mut` (returning `Option`) and document that `outgoing_flash`/`outgoing_flash_mut` panic when no `FlashHandler` is installed. The new trait methods have default bodies so adding them is not a breaking change for downstream implementors of `FlashDepotExt`.

## Testing

`cargo test -p salvo-flash --lib` (55) and `cargo test -p salvo-jwt-auth --features oidc --lib` (62) pass.

### DIFF
--- a/crates/flash/src/lib.rs
+++ b/crates/flash/src/lib.rs
@@ -277,10 +277,20 @@ pub trait FlashDepotExt {
     fn incoming_flash(&mut self) -> Option<&Flash>;
     /// Get outgoing flash, or `None` if no [`FlashHandler`] is installed on the
     /// route (so the outgoing flash was never initialized).
-    fn try_outgoing_flash(&self) -> Option<&Flash>;
+    ///
+    /// Defaulted so that adding it is not a breaking change for existing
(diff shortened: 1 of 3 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

Issue #1612, opened 2026-06-24
Panic: only one wildcard param is allowed and it must be the last one


treadful:
**Describe the bug**

Panic when making a request with an unhandled HTTP method ~~with a specific regex param~~.

```
thread 'tokio-rt-worker' (1024352) panicked at /home/treadful/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/salvo_core-0.93.0/src/routing/path_params.rs:46:17:
only one wildcard param is allowed and it must be the last one
```

~~This does not seem occur without the regex `PathFilter`.~~

**To Reproduce**

Minimal repro repo: https://github.com/treadful/salvo-greedy-panic

1) Run the server `cargo run`
2) Make a HEAD request `curl -I http://localhost:8000/b33f/subpath.txt`

**Expected behavior**

Not panic.

treadful:
My fault, the regex `PathFilter` was irrelevant. Removed.

PR #1632, opened 2026-07-01
perf(core/routing): roll back path params by truncation instead of cloning

chrislearn:
## What

`Router::detect` snapshotted the entire `PathParams` map (an `IndexMap<String, String>`) before descending into each child router, and restored it by **cloning again** on every mismatch:

```rust
let original_params = path_state.params.clone();
for child in &self.routers {
    if let Some(dm) = child.detect(...).await { return Some(dm); }
    else {
        path_state.cursor = original_cursor;
        path_state.params = original_params.clone(); // deep-copies every key/value String
    }
}
```

For a route tree of depth `d` with `b` siblings per level, matching a **single request** deep-copies every captured key/value `String` on the order of `O(b*d)` times. This is the most frequently executed allocation on the routing hot path.

## How

Matching only ever *appends* params via `PathParams::insert` (verified: the only mutation sites in `filters/path.rs` are `insert`; there is no `remove`/`clear`). So the pre-descent state is fully described by `(len, greedy)`. This PR replaces the clone with an `O(1)` snapshot/rollback pair:

```rust
pub(crate) fn snapshot(&self) -> (usize, bool) { (self.inner.len(), self.greedy) }
pub(crate) fn rollback(&mut self, (len, greedy): (usize, bool)) {
    self.inner.truncate(len);
    self.greedy = greedy;
}
```

This mirrors how `matched_parts` is **already** rolled back with `truncate` in the very same loop, so it's consistent with existing precedent. The `greedy` flag is restored too, so wildcard rollback (issue #1612 path) keeps working.

## Correctness

- No public API / behavior change — the snapshot token is `pub(crate)`.
- All existing routing tests pass, including `test_router_detect_method_mismatch_wildcard_sibling`, which exercises the wildcard/`greedy` rollback branch.
- Adds `test_router_detect_sibling_param_rollback`: a sibling captures a named param, fails on a deeper segment, and the next sibling wins — asserting the failed sibling's capture does not leak into the final params.
- Verified green with and without the `matched-path` feature; `clippy` clean.

Found as part of a broader performance review of the routing hot path.

### DIFF
--- a/crates/core/src/routing/filters/path.rs
+++ b/crates/core/src/routing/filters/path.rs
@@ -311,7 +311,7 @@ impl PathWisp for CharsWisp {
         }
         if count >= self.min_width {
             state.forward(byte_len);
-            state.params.insert(&self.name, value);
+            state.params.insert_tracked(&self.name, value);
             #[cfg(feature = "matched-path")]
             push_named_part(&mut state.matched_parts, &self.name);
             true
@@ -446,7 +446,7 @@ impl PathWisp for CombWisp {
             let mut matched_part = String::new();
             for name in self.names.iter().take(take_count) {
                 if let Some(value) = caps.name(name) {
-                    state.params.insert(name, value.as_str().to_owned());
+                    state.params.insert_tracked(name, value.as_str().to_owned());
                     #[cfg(feature = "matched-path")]
                     {
                         if value.start() > start {
@@ -509,7 +509,7 @@ impl PathWisp for CombWisp {
                 if let Some(cap) = cap {
                     let cap = cap.as_str().to_owned();
                     state.forward(cap.len());
-                    state.params.insert(wild_name, cap);
+                    state.params.insert_tracked(wild_name, cap);
                     #[cfg(feature = "matched-path")]
                     push_named_part(&mut state.matched_parts, wild_name);
                     true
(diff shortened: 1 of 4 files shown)


chrislearn:
Good catch on the in-place overwrite case, @codex — you're right that truncate alone couldn't undo an ancestor param that a failed descendant reuses (`{id}` -> `{id}/profile` matching `/alice/edit`).

Fixed in 8ed8f7b0: `insert` now records `(index, old_value)` in a small rollback log whenever it overwrites an existing entry, and `rollback` replays that log (newest-first, before truncating) to restore clobbered ancestor values, then drops the appended entries. The common path (distinct names) never overwrites, so the log stays empty and the O(1) fast path is unchanged. The log is excluded from `PathParams`' public `Debug`/`Eq` identity as transient state.

Added `test_router_detect_overwritten_param_rollback`, which reproduces your exact scenario and fails under the previous truncate-only rollback.

@codex review

chrislearn:
Addressed three review follow-ups in the latest commit:

1. **Public `insert` back to constant space.** Logging lived in the public API while `snapshot`/`rollback` are crate-private, so a user repeatedly overwriting the same key via `req.params_mut()` grew `rollback_log` with nothing draining it. Entry points are now split: public `insert` records nothing; route matching uses the crate-private `insert_tracked` (all wisp callsites switched). Belt-and-braces: `seal()` clears the log when a successful match hands params to the `Request`, so entries logged along the winning branch don't linger for the request lifetime either.

2. **`rollback` no longer silently skips a missing index** — per the invariant (logged entries replay before `truncate`) that branch is unreachable, and a `debug_assert!` now surfaces any future log/map inconsistency instead of masking it (same pattern as the wildcard guard in `insert_inner`).

3. **Doc claim corrected**: O(1) on the common append-only path, O(k) replay when k overwrites must be undone — not "O(1) amortized".

New unit tests cover all three: constant-space public inserts, tracked-insert logging + `seal`, and overwrite rollback. 212 lib tests green under both default and `matched-path` features.

@codex review

PR #1633, opened 2026-07-01
perf(core): pass-through hasher for Depot's type-keyed storage

chrislearn:
## What

`Depot`'s private `typed: HashMap<TypeId, TypedEntry>` used the default SipHash. A `TypeId` is already a unique, well-distributed 128-bit identifier produced by the compiler, and type ids are never attacker-controlled — so the DoS-resistant hasher is pure overhead on a lookup that middleware (auth, session, jwt, …) performs on every request via `get_typed`/`insert_typed`.

## How

A small pass-through `TypeIdHasher` that forwards the id bits directly via `write_u64`/`write_u128` — the same technique `http::Extensions` uses for its type-keyed map. A byte-fold fallback in `write` keeps the hasher correct if `TypeId`'s internal `Hash` impl ever changes (no `unreachable!` panic risk).

- **Zero new dependencies.**
- The map is private (`inner()` exposes only the string-keyed map), so **no API change**.
- The string-keyed `named` map intentionally keeps SipHash: its keys are arbitrary user strings.

## Verification

- All 207 `salvo_core` lib tests pass; clippy clean.
- Depot-specific tests (`test_depot_typed`, `test_depot_named_and_typed_are_separate`, …) green.

Found as part of a performance review of per-request hot paths (P4 in the review).

### DIFF
--- a/crates/core/src/depot.rs
+++ b/crates/core/src/depot.rs
@@ -1,6 +1,7 @@
 use std::any::{Any, TypeId, type_name};
 use std::collections::HashMap;
 use std::fmt::{self, Debug, Formatter};
+use std::hash::{BuildHasherDefault, Hasher};
 
 /// Store temporary data for the current request.
 ///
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

PR #1634, opened 2026-07-01
perf(serve-static): derive canonical roots lazily at acceptance time

chrislearn:
## What (final design, after three review rounds)

`StaticDir` canonicalized **every configured root upfront on every request** — filesystem syscalls paid before knowing whether any candidate file even exists.

This PR moves root canonicalization to **acceptance time, lazily**:

- The root is canonicalized only when a candidate actually exists on disk (`symlink_metadata` succeeded), right before the containment check.
- Default-file and compressed-variant probes reuse the canonical root freshly derived earlier in the *same request* (`trusted_canonical_root`), as before.
- **Misses skip root canonicalization entirely**: 404s, dot-file rejections, and roots whose candidate doesn't exist (multi-root setups) pay zero root syscalls. The hit path costs the same as before.

## Why not a cross-request cache (design history)

Earlier iterations of this PR cached canonical roots across requests (TTL + single-flight + mismatch-healing). Codex review demonstrated the approach is **unsound on the acceptance side**: when a symlinked root is retargeted so the old canonical prefix is *broader* than the new root (e.g. `current` moved from `/srv/www` into `/srv/www/release-b`), an escape like `current/leak → /srv/www/secret.txt` still canonicalizes under the stale prefix and is **accepted** — no mismatch fires, so no heal/TTL can catch it. The containment check is only sound against the root's *current* canonical path, and a successful prefix match can't be proven fresh without re-canonicalizing. Hence: no cache, lazy derivation instead.

## Security

- Escape checks always run against the root's current canonical path — identical guarantees to `main`, including mid-request symlink retargets.
- New regression test `test_resolve_rejects_escape_after_root_narrowing_retarget` encodes the Codex P1 scenario (fails against the cached design).
- `test_resolve_uses_fresh_root_after_symlink_retarget` covers deploy swaps serving the new release immediately; `test_resolve_picks_up_late_created_root` covers late-created roots.

## Verification

All `salvo-serve-static` tests pass (unix-only symlink tests run on Linux CI); clippy clean; `cargo fmt` applied.

### DIFF
--- a/crates/serve-static/src/dir.rs
+++ b/crates/serve-static/src/dir.rs
@@ -6,9 +6,7 @@ use std::fmt::{self, Debug, Display, Formatter, Write};
 use std::fs::Metadata;
 use std::path::{Component, Path, PathBuf};
 use std::str::FromStr;
-use std::sync::atomic::{AtomicBool, Ordering};
-use std::sync::{Arc, PoisonError, RwLock};
-use std::time::{Duration, Instant, SystemTime};
+use std::time::SystemTime;
 
 use salvo_core::fs::NamedFile;
 use salvo_core::handler::Handler;
@@ -181,13 +179,6 @@ pub struct StaticDir {
     pub defaults: Vec<String>,
     /// Fallback file to serve when requested file isn't found
     pub fallback: Option<String>,
-    /// Cache of canonicalized roots, revalidated against the current (public,
-    /// mutable) `roots` value on every request and re-resolved after a short TTL.
-    /// See [`StaticDir::canonical_roots`].
-    canonical_cache: RwLock<Option<Arc<CanonicalCache>>>,
-    /// Claimed by the single request that re-resolves an expired cache, so a TTL
-    /// expiry under concurrent traffic does not stampede `canonicalize` calls.
-    canonical_refreshing: AtomicBool,
 }
 impl Debug for StaticDir {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
@@ -223,8 +214,6 @@ impl StaticDir {
             compressed_variations,
             defaults: vec![],
             fallback: None,
-            canonical_cache: RwLock::new(None),
-            canonical_refreshing: AtomicBool::new(false),
         }
     }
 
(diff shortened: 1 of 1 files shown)


chrislearn:
Right — the `current -> release-N` layout would 404 indefinitely after a retarget. Fixed with a TTL rather than per-request revalidation of symlinked roots: `read_link` on the root alone can't detect a retargeted symlink in a *middle* path component, so any correct revalidation is a full `canonicalize` — which is exactly what we're caching.

Cached roots now expire after `CANONICAL_ROOTS_TTL` (1s). A retarget is picked up within a second, and the steady-state cost is bounded at one re-resolution per root per second regardless of request rate — at 1k RPS that's still a 1000× reduction over the old per-request behavior.

Added `test_canonical_roots_reresolved_after_ttl` plus a unix end-to-end test reproducing your exact scenario (`current -> release-a` retargeted to `release-b`, old canonical rejected, new one served).

@codex review

chrislearn:
Fixed — expired refreshes are now single-flight. One request claims the refresh via an atomic CAS; the others keep serving the previous cache when it was built from the same `roots` (staleness bounded by the refresh duration). Two edge cases handled deliberately:

- A cache built from **different** roots is never served stale — those requests resolve immediately without stealing the claim (duplicate work only in the rare first-request / roots-mutation window, where correctness beats coalescing).
- The claim is released by a drop guard, so a request cancelled mid-`canonicalize` (client disconnect) can't wedge refreshes permanently.

`test_canonical_roots_stale_while_revalidate` covers all three branches: stale served while the claim is held, roots-change bypassing staleness, and normal refresh after release.

@codex review

chrislearn:
You're right — during the TTL window the stale canonical root didn't just serve old content, it 404'd everything: resolution followed the retargeted symlink to `release-b` and then rejected it against the cached `release-a` prefix.

Fixed with mismatch-triggered healing rather than serving from the cached canonical target (which would silently change what `exclude_filters` see, and would serve the *old* release instead of the new one). On a canonical-prefix mismatch, the root is re-canonicalized once — one syscall on the short root path — to distinguish the two causes:

- root unchanged / target still outside → genuine escape attempt, rejected exactly as before;
- root retargeted → serve against the fresh root **immediately** and patch the cached entry in place (same-generation check, `resolved_at` preserved) so later requests don't repeat the extra syscall.

Deploy swaps now have a zero-404 window and serve the new release at once; the cost is one extra `canonicalize` only on the rare mismatch path (escape attempts included, which bounds any adversarial amplification at 2× per malicious request).

New unix test `test_resolve_heals_retargeted_symlink_root_immediately` covers: heal-and-serve without TTL expiry, in-place cache patch, and escape rejection against the healed root.

@codex review

chrislearn:
This P1 is decisive — thank you. The narrowing-retarget scenario shows the cache is unsound on the **acceptance** side: a successful `starts_with` against a stale broader prefix accepts escapes, and no mismatch-triggered healing or TTL can catch a *successful* match. A prefix match simply cannot be proven fresh without re-canonicalizing the root, which is the very syscall the cache existed to avoid.

So I've removed the caching design entirely (TTL, single-flight, healing — all of it) and replaced it with **lazy acceptance-time derivation**: the root is canonicalized only when a candidate actually exists, right before the containment check; child probes (default files / compressed variants) reuse the same request's freshly derived root exactly as before. Security semantics are now identical to `main` — every acceptance is checked against the root's current canonical path — while misses (404s, dot-file rejections, non-matching roots in multi-root setups) skip root canonicalization entirely, which preserves a real, if smaller, win over the original upfront-loop.

Your scenario is now a regression test: `test_resolve_rejects_escape_after_root_narrowing_retarget` (`current` retargeted from `www` into `www/release-b`, escape `leak → www/secret.txt` rejected; fails against the cached design).

@codex review

PR #1635, opened 2026-07-01
fix(core)!: let each Scribe own its body semantics; Json render now replaces

chrislearn:
## Design principle (from maintainer review)

Whether a render **replaces** or **appends** is not a global property of `Response::render` — it belongs to each `Scribe`:

- A footer writer appends by nature — clearing the body would be wrong.
- An NDJSON emitter appends one record per call — append *is* its format.
- A JSON document must replace — two concatenated documents are invalid JSON.

Earlier iterations of this PR added a global debug warning whenever a render appended onto existing bytes. That approach contradicts the principle above: it would false-positive on legitimately appending scribes. It has been removed.

## What this PR now does

1. **`Json::render` replaces** previously buffered body bytes (`None`/`Error`/`Once`/`Chunks`) instead of appending invalid output. Debug builds log a warning when non-empty content is discarded — that usually indicates a missing early-`return` after an error render, and with replace semantics the earlier content would otherwise vanish silently. Streaming body kinds are left untouched with an error log, mirroring `write_body`'s behavior.
2. **Text-like scribes keep append semantics** (`&str`, `String`, `Text`): concatenated text is valid text, and incremental writing is a legitimate pattern.
3. **`Scribe` trait docs** state the design rule ("pick the semantics that make the final body valid in your format") with the footer/NDJSON/JSON examples; `Response::render` and `write_body` docs updated to match. `Response::render` itself is pure delegation again; the `Writer` blanket impls are back to calling `Scribe::render` directly (zero diff from `main` in `writing.rs` apart from docs).

## Breaking change

Rendering `Json` after other body bytes were written now **replaces** them (previously: concatenated). Since concatenated JSON was invalid output, any code affected by this was already producing broken responses.

## Verification

- New tests: `test_json_render_replaces_previous_body` (double render yields the last, valid document), `test_json_render_does_not_clobber_stream_body`.
- All 209 `salvo_core` lib tests pass; clippy clean.
- Downstream: `salvo-flash` (55), `salvo-cache` (70), `salvo-csrf` (30) all green. (`salvo-oapi` has 2 pre-existing failures reproducible on `main`; `salvo_extra` hits a pre-existing `rustls`/`aws_lc_rs` feature-combination compile issue in `conn/rustls/config.rs`, unrelated to this diff.)

### DIFF
--- a/crates/core/src/http/response.rs
+++ b/crates/core/src/http/response.rs
@@ -482,7 +482,10 @@ impl Response {
             }
             let before = committed_len(&self.body);
             scribe.render(self);
+            // `before > 0`: appending to an *empty* committed body yields exactly the
+            // new render's bytes, so nothing is corrupted and no warning is needed.
             if let (Some(before), Some(after)) = (before, committed_len(&self.body))
+                && before > 0
                 && after > before
             {
                 tracing::warn!(
--- a/crates/core/src/http/response.rs
+++ b/crates/core/src/http/response.rs
@@ -446,6 +446,14 @@ impl Response {
     /// `Scribe` impl is responsible for converting the error into a `5xx` response
     /// instead of panicking or returning an error here.
     ///
+    /// **Note**: the built-in `Scribe` implementations write through
+    /// [`write_body`](Response::write_body), which *appends* to any body bytes
+    /// already present — it does not overwrite them. Calling `render` twice on the
+    /// same response therefore concatenates both outputs, which for `Json` produces
+    /// invalid JSON. Debug builds log a warning when this happens. To replace an
+    /// already-written body, set it explicitly via
+    /// [`replace_body`](Response::replace_body) or [`body`](Response::body) first.
+    ///
     /// # Example
     ///
     /// ```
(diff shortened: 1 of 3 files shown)


chrislearn:
Good catch — the blanket `Writer for P: Scribe` / `Option<P>` impls called `Scribe::render` directly and bypassed the diagnostic for returned scribes. Fixed: both impls now route through `Response::render`.

Doing that surfaced a false-positive your scenario implies but didn't mention: `Result<(), E>` handlers (`res.render("a"); Ok(())`) render the `()` scribe after legitimately writing the body, so a "bytes already present" pre-check would warn on the most idiomatic salvo pattern. The diagnostic now snapshots committed body size before the scribe runs and warns only when the render **actually appended** onto pre-existing bytes — `()` writes nothing and stays silent, while `res.render("a"); return "b"` now warns as you suggested.

@codex review

chrislearn:
Fixed — the warning now additionally requires `before > 0`: appending to an empty committed body produces exactly the new render's bytes, so it stays silent. Only a render that appends onto non-empty pre-existing content warns.

@codex review

chrislearn:
@codex review

chrislearn:
Reworked per maintainer feedback: replace-vs-append is each `Scribe`'s own decision, not a global rule of `render` — a footer writer appends, an NDJSON emitter appends, a JSON document replaces. The global append-detection warning from earlier iterations is removed (it would false-positive on legitimately appending scribes), `Response::render` is pure delegation again, and `Json::render` now carries the replace semantics itself (with a debug warning when it discards non-empty content, and streaming bodies left untouched like `write_body`). Text-like scribes keep appending. See the updated PR description for details.

@codex review

chrislearn:
@codex review

PR #1637, opened 2026-07-01
perf(compression): match content types without full MIME parsing

chrislearn:
## What

`Compression::negotiate` ran the full `Mime` parser on every response's `Content-Type` header, but the subsequent check only compares the `type/subtype` essence against the configured list — parameter parsing, quoting, and validation are wasted work on a per-response hot path.

## How

Extract the essence from the raw header (`split(';')` + `split_once('/')`) and compare case-insensitively per RFC 2045.

Semantics preserved:
- No `/`, or empty type/subtype → no match (the old parser rejected these too — new test `test_content_types_malformed_not_compress`).
- Parameters ignored as before; mixed-case headers (`TEXT/HTML; Charset=UTF-8`) still match `text/html` (new test `test_content_types_match_is_case_insensitive_and_ignores_params` — the old path lowercased via the parser, the new path uses `eq_ignore_ascii_case`).

## Verification

All 65 `salvo-compression` lib tests pass (63 existing + 2 new); clippy clean.

Found as part of a performance review of per-request hot paths (P10 in the review).

### DIFF
--- a/crates/compression/src/lib.rs
+++ b/crates/compression/src/lib.rs
@@ -370,13 +370,19 @@ impl Compression {
             }
             // Only the `type/subtype` essence is compared, so extract it from the raw
             // header instead of running the full `Mime` parser (parameters, quoting,
-            // validation) on every response. MIME types compare case-insensitively.
+            // validation) on every response. MIME types compare case-insensitively,
+            // and — matching `Mime::subtype()`, which excludes the structured suffix —
+            // the `+suffix` is ignored so `image/svg+xml` matches a configured
+            // `image/svg+xml` entry (whose `subtype()` is just `svg`).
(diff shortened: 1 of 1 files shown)


chrislearn:
Confirmed and fixed — wrote the regression test first (`image/svg+xml` against the default list) and verified it failed on the previous commit: `Mime::subtype()` is indeed suffix-less (`svg`), so the raw comparison broke SVG/RSS compression.

The fast path now strips the `+suffix` from the raw subtype before comparing (`rsplit_once('+')`, matching the crate's last-suffix semantics), which restores the old parser's suffix-insensitive matching on both sides. All 66 tests pass including the new regression test.

@codex review

PR #1638, opened 2026-07-01
fix(rate-limiter): deprecate ForwardedHeaderIssuer

chrislearn:
## Why

`ForwardedHeaderIssuer` unconditionally trusts `X-Forwarded-For` / `X-Real-IP`. Any client that can reach the application directly — a developer port-forward, a misrouted firewall rule, a brief security-group regression — can spoof arbitrary IPs and bypass rate limiting entirely. The module already ships the safe alternative, `TrustedProxyIssuer`, which honours those headers only when the request actually arrived from a configured proxy address.

The docs already carried strong warnings, but a warning in prose doesn't reach someone skimming autocomplete; a `#[deprecated]` note does.

## What

- `#[deprecated(since = "0.94.0")]` on `ForwardedHeaderIssuer` with a note explaining the spoofing risk and pointing at `TrustedProxyIssuer` / `RemoteIpIssuer`, plus a one-line migration hint.
- Fixed `RemoteIpIssuer`'s docs, which still **recommended** `ForwardedHeaderIssuer` for proxy deployments — guidance that predated `TrustedProxyIssuer` and steered users toward the spoofable variant.
- Updated the crate-level issuer list to lead with the safe options.

No behavior change; the type keeps working under `#[allow(deprecated)]`-gated impls.

## Verification

53 lib tests pass (`--all-features`); clippy clean; doctests compile.

Found as part of a security review (footgun-reduction item: rate-limiter bypass via forged forwarded headers).

### DIFF
--- a/crates/rate-limiter/src/lib.rs
+++ b/crates/rate-limiter/src/lib.rs
@@ -18,11 +18,12 @@
 //!
 //! ## Issuers
 //! - [`RemoteIpIssuer`]: Identifies clients by their direct connection IP
-//! - [`ForwardedHeaderIssuer`]: Unconditionally trusts `X-Forwarded-For` / `X-Real-IP` (legacy;
-//!   **use only when the application is unreachable except through a header-rewriting proxy**)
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

PR #1639, opened 2026-07-02
fix(examples): pin sea-orm to rc.40 to unblock CI

chrislearn:
## Why

`sea-orm 2.0.0-rc.41` (published today) fails to compile — its `sea_schema_shim` implements trait methods whose returned futures lack the required `Send` bound (`E0053`, in `sea-orm-2.0.0-rc.41/src/database/sea_schema_shim.rs`). The `db-postgres-sea-orm` example uses the floating requirement `"2.0.0-rc"`, so every fresh **check examples** CI run now resolves the broken release and fails — currently breaking CI on all open PRs (e.g. #1637, #1638 both hit it with identical errors, while earlier runs today passed on rc.40).

## What

Pin the example to `=2.0.0-rc.40`. Verified locally: `cargo check -p salvo-postgres-seaorm` compiles clean against rc.40.

Revert to a floating requirement once upstream ships a fixed rc (tracked upstream in SeaQL/sea-orm).

### DIFF
--- a/examples/db-postgres-sea-orm/Cargo.toml
+++ b/examples/db-postgres-sea-orm/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2024"
 [dependencies]
 migration = { path = "migration" }
 anyhow.workspace = true
-sea-orm = { version = "2.0.0-rc", features = ["with-chrono", "sqlx-postgres", "macros", "with-uuid", "with-time", "runtime-tokio", "runtime-tokio-native-tls", "runtime-tokio-rustls", "schema-sync", "entity-registry"] }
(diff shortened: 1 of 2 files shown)


chrislearn:
Right — the nested `migration` crate's `sea-orm-migration = "2.0.0-rc.20"` resolved to rc.41, whose `sea-orm ~rc.41` requirement clashed with the `=rc.40` pin and made resolution fail outright. Pinned `sea-orm-migration = "=2.0.0-rc.40"` as well; verified with a fresh `Cargo.lock` that both crates now resolve to rc.40.

@codex review

PR #1641, opened 2026-07-05
docs(core): clarify three API footguns; implement StatusError::source

chrislearn:
Documentation-only PR addressing three spots where the docs made misuse easy (H2/M2/M5 from the API-ergonomics review):

1. **`Request::param` swallows parse errors** — it returns `None` for both "key missing" and "value present but failed to parse", and its doc was a single line that said neither. Now states the merge explicitly, points at `try_param` (documented as the `?`-friendly variant that distinguishes the two via `ParseError`) and at `Extractible` for strongly-typed extraction. `query()`/`form()` already carried this note; `param()` was the gap.

2. **`FlowCtrl::skip_rest` does not set the ceased flag** — the documented around-middleware pattern (`call_next` then check `is_ceased` before post-processing) silently keeps running post-processing when the chain was ended via `skip_rest` or the automatic terminal-response skip. `skip_rest` and `is_ceased` now spell out exactly which actions set the flag and when to use `cease` instead.

3. **`StatusError::cause` vs `origin`** — both fields were documented only as "similar to the other field", forcing a guess on every custom error. Each now states its decision rule: `cause` for standard error reporting (logging, `source` chains), `origin` when a downstream handler/catcher needs the concrete value back via `downcast_origin` (type need not implement `Error`).

No behavior change. 215 lib tests pass; no new rustdoc warnings.

### DIFF
--- a/crates/core/src/http/errors/status_error.rs
+++ b/crates/core/src/http/errors/status_error.rs
@@ -43,9 +43,20 @@ pub struct StatusError {
     pub brief: String,
     /// Detailed information about the HTTP error.
     pub detail: Option<String>,
-    /// Cause of the HTTP error. Similar to the `origin` field, but using [`std::error::Error`].
+    /// The underlying error, as a [`std::error::Error`] trait object.
+    ///
+    /// Choose `cause` when you have a real error value and want it reported
+    /// through standard error tooling (logging, `source` chains). Choose
+    /// [`origin`](Self::origin) instead when a later handler or catcher needs the
+    /// concrete value back via [`StatusError::downcast_origin`].
     pub cause: Option<Box<dyn StdError + Sync + Send + 'static>>,
-    /// Origin of the HTTP error. Similar to the `cause` field, but using [`std::any::Any`].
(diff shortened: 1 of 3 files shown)


chrislearn:
Good catch — the docs promised `source`-chain reporting that the default `source()` impl (always `None`) didn't deliver; the cause only surfaced via `Display`. Fixed by making the claim true rather than weakening it: `StatusError` now implements `source()` returning the stored cause, so standard error reporters and chain walkers can traverse it. Test covers both populated and empty cases.

The PR is no longer strictly docs-only; title updated accordingly.

@codex review

chrislearn:
Clarified the naming in the field docs: `cause` is the concept (and this field's pre-existing public name); `source()` is the std API that reports it — `Error::source` replaced the deprecated `Error::cause` method in Rust 1.30 (RFC 2504) because the old signature's missing `'static` bound prevented downcasting. thiserror's `#[source]` expands to exactly this pattern. Renaming the public `cause` field to `source` for ecosystem symmetry would be a breaking change with little payoff; keeping the field name and documenting the relationship is the pragmatic call.

@codex review

PR #1647, opened 2026-07-05
perf: borrow undecoded path segments

chrislearn:
## Summary
- return `Cow<'_, str>` from `decode_url_path`, borrowing plain path segments when no percent encoding is present
- store `PathState` as `Cow<'_, str>` plus raw segment ranges, so callers can choose borrowed or owned path storage explicitly
- deprecate ambiguous `PathState::new` in favor of `from_borrowed_path`, `from_owned_path`, and `from_cow_path`
- keep service dispatch sound by moving an owned request path into `PathState`, while direct/benchmark routing paths can use borrowed storage
- add coverage for borrowed/plain paths, owned/decoded paths, and UTF-8 byte-boundary handling

## Notes
- request dispatch still copies the request path because `Router::detect` receives `&mut Request`; borrowing from the same request would conflict with the mutable borrow
- callers that already have a stable path string can now avoid that copy with `PathState::from_borrowed_path`

## Tests
- `cargo test -p salvo_core -p salvo-serve-static --quiet`
- `cargo check -p salvo_core --features matched-path --quiet`
- `cargo bench -p salvo_core --bench router_detect --no-run --quiet`

### DIFF
--- a/crates/core/benches/router_detect.rs
+++ b/crates/core/benches/router_detect.rs
@@ -24,9 +24,10 @@ fn bench_detect(c: &mut Criterion, name: &str, router: &Router, url: &str) {
         .build()
         .expect("build runtime");
     let mut req = TestClient::get(url).build();
+    let path = req.uri().path().to_owned();
     c.bench_function(name, |b| {
         b.iter(|| {
-            let mut state = PathState::new(req.uri().path());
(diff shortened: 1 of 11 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

PR #1651, opened 2026-07-05
perf: inline tcp idle timeout polling

chrislearn:
## Summary
- remove the per-connection idle-timeout background task created by `FlexFactory::build`
- track the last TCP activity timestamp and drive the idle timer from `FlexFusewire::fused`
- keep frame and TLS timeout behavior unchanged, including their existing cancellation-token state machine

## Notes
- activity updates `last_activity` before notifying active `fused()` waiters; the idle wait path re-checks the timestamp after waking so a raced activity event cannot fuse the connection early

## Tests
- `cargo test -p salvo_core fuse::flex::tests`
- `cargo test -p salvo_core`


chrislearn:
@codex review

PR #1672, opened 2026-07-14
Skip duplicate WebTransport response writes

chrislearn:
## Summary

- return the recovered HTTP/3 connection immediately after an accepted WebTransport session ends
- keep the accepted CONNECT stream out of the normal HTTP/3 response writer
- preserve the existing response path for WebTransport requests that were not accepted

## Root cause

Follow-up to #1671 and its P2 review. `WebTransportSession::accept()` sends the successful CONNECT response before the session is stored in response extensions. After recovering the session, `process_web_transport` reused its CONNECT stream in the common response path, which attempted to send another response and then finish the stream.

## Impact

Accepted WebTransport sessions now restore the server connection without sending duplicate response headers, body frames, trailers, or a second explicit finish through the normal response path.

## Validation

- `cargo test -p salvo_core --features quinn` (249 unit tests and 57 doctests passed)
- `cargo check -p salvo_core --features quinn`
- `cargo check --manifest-path examples/webtransport/Cargo.toml`
- `cargo check --manifest-path examples/webtransport-acme-http01/Cargo.toml`
- `cargo clippy -p salvo_core --features quinn --all-targets -- -D warnings -A clippy::useless_borrows_in_formatting -A clippy::question_mark -A clippy::redundant_clone`
- `cargo fmt --all -- --check`
- `git diff --check`

The three allowed Clippy lint classes are pre-existing warnings on `main` and are unrelated to this change.

### DIFF
--- a/crates/core/src/conn/quinn/builder.rs
+++ b/crates/core/src/conn/quinn/builder.rs
@@ -268,36 +268,34 @@ async fn process_web_transport(
         .await
         .map_err(|e| IoError::other(format!("failed to call hyper service : {e}")))?;
 
-    let conn;
-    let stream;
     if let Some(session) = take_unique_arc_extension::<
         WebTransportSession<salvo_http3::quinn::Connection, Bytes>,
     >(response.extensions_mut(), "WebTransport session")?
     {
-        let (server_conn, connect_stream) = session.split();
-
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

PR #1682, opened 2026-07-29
Document core HTTP types and validate ranges

chrislearn:
## Summary

- expand `SecureMaxSize` and global request-size-limit documentation with precedence, middleware ordering, caching, and direct-stream caveats
- document `HttpRange` normalization and parsing behavior with executable examples
- clarify `BytesFrame`, `ReqBody`, `ResBody`, `BodySender`, and `BodyReceiver` semantics
- replace stale `BodySender` links and explain data/trailer channel shutdown behavior
- reject range members with extra hyphens and suffix ranges against an empty representation

## Why

Several public HTTP types had only terse labels, while the `BodySender` comments referenced APIs that no longer exist. During review, the documented malformed-range contract also exposed two parser gaps: trailing hyphen-separated components were ignored, and a suffix range over an empty representation produced a zero-length range.

The updated documentation makes the security boundary and body/range APIs easier to use correctly, while the parser changes keep its validation behavior consistent with that contract.

## Impact

Malformed byte ranges such as `bytes=0-1-2` and suffix ranges against an empty representation now return `ParseError::InvalidRange`. Valid range behavior and the remaining HTTP APIs are unchanged.

## Validation

- `cargo test -p salvo_core --lib` — 248 passed
- `cargo test -p salvo_core --doc` — 59 passed, 29 ignored
- `cargo fmt -p salvo_core -- --check`
- `git diff --check`
- `cargo clippy -p salvo_core --all-targets -- -D warnings` with four unrelated pre-existing lint categories allowed
- `cargo doc -p salvo_core --no-deps` — completed; reports eight pre-existing warnings outside the edited HTTP files

### DIFF
--- a/crates/core/src/http/body.rs
+++ b/crates/core/src/http/body.rs
@@ -64,7 +64,9 @@ use crate::http::HeaderMap;
 /// A frame contains either body data or trailer headers. Use [`BytesFrame::data`]
 /// to create a data frame, and [`BytesFrame::into_data`] or
 /// [`BytesFrame::into_trailers`] to distinguish the two frame kinds without
-/// losing the original frame on a mismatch.
+/// losing the original frame on a mismatch. Converting a `BytesFrame` directly
+/// into [`Bytes`] returns an empty buffer for a trailers frame, so use
+/// [`BytesFrame::into_data`] when that distinction matters.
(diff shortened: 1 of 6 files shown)


chrislearn:
@codex review

PR #1688, opened 2026-08-06
Add OpenAPI 3.2 object model and version-aware generation

### LINKED ISSUE #1685: Add OpenAPI 3.2 support to salvo-oapi
## Problem

`salvo-oapi` currently emits and accepts OpenAPI 3.1.x documents only. OpenAPI 3.2.0 is backward-compatible with 3.1, but it adds a new document version and several optional fields and objects that are not represented by the current model.

References:

- [OpenAPI 3.2.0 specification](https://spec.openapis.org/oas/v3.2.0.html)
- [Official 3.1 to 3.2 upgrade guide](https://learn.openapis.org/upgrading/v3.1-to-v3.2.html)

## Proposed rollout

Keep OpenAPI 3.1 as the default initially. Add 3.2 support incrementally so existing generated documents and downstream users are not changed unexpectedly.

### Phase 1: baseline 3.2 compatibility

- [x] Add `OpenApiVersion::Version3_2` and accept `3.2.x` during deserialization
- [x] Add the root `$self` field and builder support
- [x] Add serialization/deserialization and regression tests for both 3.1 and 3.2
- [x] Document that callers must opt in to 3.2 explicitly

### Phase 2: OpenAPI 3.2 object model

- [x] Tag Object: `summary`, `parent`, and `kind`
- [x] Server Object: `name`
- [x] Path Item Object: `query` and `additionalOperations`
- [x] Parameter Object: `in: querystring`, `style: cookie`, and updated `allowReserved` handling
- [x] Media Type Object: `itemSchema`, `prefixEncoding`, and `itemEncoding`
- [x] Encoding Object: nested `encoding`, `prefixEncoding`, and `itemEncoding`
- [x] Example Object: `dataValue` and `serializedValue`
- [x] Response Object: optional `description` and `summary`
- [x] Components Object: reusable `mediaTypes`
- [x] Discriminator Object: `defaultMapping`
- [x] XML Object: `nodeType`
- [x] Security Scheme Object: `deprecated` and `oauth2MetadataUrl`
- [x] OAuth Flows Object: Device Authorization flow and `deviceAuthorizationUrl`
- [x] Review Security Requirement URI references and add conformance tests

### Phase 3: generation, tooling, and compatibility

- [x] Teach route discovery to emit `QUERY` and custom methods via `additionalOperations`
- [x] Extend endpoint/parameter macros for new parameter locations and styles
- [x] Decide how reusable Media Type references should be exposed without unnecessary public API breakage
- [x] Add 3.2 examples and round-trip fixtures
- [x] Smoke-test Swagger UI, Scalar, RapiDoc, and ReDoc rendering with a 3.2 document; pin known-compatible assets where appropriate
- [x] Decide whether 3.2-only fields need version-aware validation when a document still declares 3.1
- [x] Address existing model gaps needed for conformance, including Info `summary` and the fuller Header Object surface
- [x] Add an `Unreleased` changelog entry when user-visible support is complete

## Compatibility notes

Several affected types are public enums (`OpenApiVersion`, `PathItemType`, `ParameterIn`, `ParameterStyle`, and `Flow`). Adding variants can break downstream exhaustive matches even though the wire-format change is additive. The implementation should call this out in release notes and avoid changing the default emitted version until an appropriate release boundary.

OpenAPI 3.2 continues to use JSON Schema Draft 2020-12 and the OpenAPI 3.1 base dialect, so this work should not require a wholesale `ToSchema` rewrite.

chrislearn:
Closes #1685.

Completes phases 2 and 3 of OpenAPI 3.2 support. Phase 1 (`OpenApiVersion::Version3_2` and the root `$self` field) landed in #1686, and QUERY method routing in salvo-core landed in #1687.

## What changed

**Phase 2 — 3.2 object model.** All thirteen checklist items:

| Object | Added |
| --- | --- |
| Tag | `summary`, `parent`, `kind` |
| Server | `name` |
| Path Item | `query` (via `PathItemType::Query`), `additionalOperations` |
| Parameter | `in: querystring`, `style: cookie`, updated `allowReserved` semantics |
| Media Type | `itemSchema`, `prefixEncoding`, `itemEncoding` |
| Encoding | nested `encoding`, `prefixEncoding`, `itemEncoding` |
| Example | `dataValue`, `serializedValue` |
| Response | `summary`; `description` now optional on input |
| Components | `mediaTypes` |
| Discriminator | `defaultMapping` |
| XML | `nodeType` (new `XmlNodeType` enum) |
| Security Scheme | `deprecated`, `oauth2MetadataUrl` |
| OAuth Flows | Device Authorization flow + `deviceAuthorizationUrl` |

Security Requirement URI references were reviewed: the model is a plain `BTreeMap<String, Vec<String>>`, so URI names and `./`-prefixed relative references already work. Added docs and a conformance test rather than code.

**Phase 3 — generation, macros, docs.** Route discovery now emits `QUERY` and custom HTTP methods (the latter via `additionalOperations`), and does so *version-aware*: since both are 3.2-only constructs, `merge_router` skips them with an actionable warning while the document still declares 3.1. This also resolves the checklist's open question about version-aware validation. Macros learned the `QueryString` location and `Cookie` style. Also closes the model gaps the issue called out — Info `summary` and the fuller Header Object surface (`required`, `deprecated`, `style`, `explode`, `example`, `examples`, `content`, extensions, plus `Header::with_content`).

Reusable Media Type references are exposed without the churn of turning every `content` map into `RefOr<Content>`: `Content` gained an optional `$ref` (`Content::from_ref`), mirroring the existing `PathItem::ref_location` convention. A `Content` carrying only that field serializes exactly as a Reference Object.

**3.1 remains the default emitted version.**

## Pre-existing defects fixed along the way

These are not 3.2 features, but the round-trip fixtures the issue asks for cannot pass without them:

1. **`PathItem` duplicated operations into `extensions`.** It flattens both the operation map and the extension map, so serde handed every unmatched key to *both* — a parsed path item re-serialized with repeated keys. `PathItem` now implements `Deserialize` explicitly and partitions keys.
2. **`clientCredentials` OAuth2 flows deserialized as `Flow::Password`.** `Flow` is untagged and those two variants have identical fields, so the first match won. Flows are now dispatched by their flow name, which additionally keeps the new device flow independent of enum declaration order.
3. **salvo-oapi could not deserialize its own output.** Many fields carried `skip_serializing_if` without a matching `default`, so any omitted optional field failed with `missing field`. Added where needed. (`Option` fields already default implicitly and were left alone — an early blanket pass also hit `Object::schema_type`, which broke `Schema`'s untagged variant discrimination; that was reverted.)

## Breaking changes

The wire-format change is purely additive, but the Rust API is not:

- New variants: `PathItemType::Query`, `ParameterIn::QueryString`, `ParameterStyle::Cookie`, `Flow::DeviceAuthorization` — these break downstream exhaustive matches.
- `SecurityScheme::MutualTls` gained a `deprecated` field, breaking struct-literal construction.
- `Header::schema` is now `Option<RefOr<Schema>>` so a header can describe its value with `content`. `Header::default()` still yields a `String` schema, so serialized output is unchanged.

All are noted in the `Unreleased` changelog entry.

## Testing

- Full workspace suite green: 55 test targets, `cargo fmt` applied, no new clippy warnings in the touched crates.
- New `crates/oapi/tests/openapi_3_2.rs` round-trip fixtures: every new field is asserted on the wire under its spec name, the document round-trips (`serialize → deserialize → serialize` is stable), and a 3.1 document is asserted unchanged.
- New `examples/oapi-3-2` was run end-to-end; the served document carries `"openapi": "3.2.0"`, `$self`, server `name`, tag `summary`/`kind`, and the `query` operation under `/search`.

## Reviewer note: documentation UI compatibility

I smoke-tested a 3.2 document against all four bundled UIs in a browser. All four load it, but only the vendored **Swagger UI v5.32.11 renders `query` operations** — Scalar, RapiDoc and ReDoc display the rest of the document and silently omit them.

I deliberately did **not** pin the Scalar/RapiDoc/ReDoc CDN URLs, which the checklist listed as an option. They currently track `latest`, and no released build of those three renders `query` yet, so pinning would lock in the same gap while forfeiting support the moment upstream ships it. Findings are recorded in the changelog and in the `OpenApi::openapi_version` docs. Happy to pin anyway if you'd rather have reproducible asset versions — that's a separate trade-off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,14 +10,62 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 - Changelog established for upcoming releases.
 - Initial opt-in OpenAPI 3.2 document support in `salvo-oapi`, including the `$self` field.
+- OpenAPI 3.2 object model in `salvo-oapi`:
+  - Tag Object `summary`, `parent` and `kind`.
+  - Server Object `name`.
+  - Path Item Object `query` (via `PathItemType::Query`) and `additionalOperations`.
+  - Parameter Object `in: querystring` (`ParameterIn::QueryString`) and `style: cookie`
+    (`ParameterStyle::Cookie`).
+  - Media Type Object `itemSchema`, `prefixEncoding` and `itemEncoding`.
+  - Encoding Object nested `encoding`, `prefixEncoding` and `itemEncoding`.
+  - Example Object `dataValue` and `serializedValue`.
+  - Response Object `summary`, and `description` is now optional on input.
+  - Components Object `mediaTypes`.
+  - Discriminator Object `defaultMapping`.
+  - XML Object `nodeType` (new `XmlNodeType` enum).
+  - Security Scheme Object `deprecated` and `oauth2MetadataUrl`.
+  - OAuth Flows Object `deviceAuthorization` (new `DeviceAuthorization` flow).
+  - Info Object `summary`.
+- Header Object gained the rest of its spec surface: `required`, `deprecated`, `style`,
+  `explode`, `example`, `examples`, `content` and extensions, plus `Header::with_content`.
+- Route discovery emits `QUERY` operations and custom HTTP methods (via
+  `additionalOperations`) when the document declares OpenAPI 3.2. Documents that still declare
+  3.1 skip such routes and log a warning explaining how to opt in.
+- `#[endpoint]` parameters accept the `QueryString` location and the `Cookie` style.
+- `Content::from_ref` / `Content::ref_location`, so a `content` map entry can reference a
+  reusable Media Type Object without changing those maps to `RefOr<Content>`.
+- `examples/oapi-3-2` demonstrates emitting a 3.2 document with a `QUERY` route.
 
 ### Changed
 
-- None yet.
+- **Breaking.** Several public enums and structs gained variants or fields, which can break
+  downstream exhaustive matches and struct literals even though the wire format change is
+  purely additive:
+  - `PathItemType::Query`, `ParameterIn::QueryString`, `ParameterStyle::Cookie` and
+    `Flow::DeviceAuthorization` are new variants.
+  - `SecurityScheme::MutualTls` gained a `deprecated` field.
+  - `Header::schema` is now `Option<RefOr<Schema>>` so a header can describe its value with
+    `content` instead. `Header::default()` still yields a `String` schema, so serialized output
+    is unchanged.
+- `OpenApi` still defaults to emitting OpenAPI 3.1; 3.2 remains opt-in via
+  `OpenApi::openapi_version`.
+
+### Notes
+
+- Documentation UI compatibility with a 3.2 document, smoke-tested against the bundled Swagger
+  UI v5.32.11 and the current CDN builds of the others: all four load the document, but only
+  Swagger UI renders `query` operations. Scalar, RapiDoc and ReDoc display the rest of the
+  document and silently omit them. Their CDN URLs are deliberately left unpinned, since no
+  released build of those three renders `query` yet.
 
 ### Fixed
 
-- None yet.
+- `PathItem` no longer duplicates operations into `extensions` when deserialized, which
+  previously made a parsed path item re-serialize with repeated keys.
+- OAuth2 flows are now resolved by their flow name instead of by shape, so a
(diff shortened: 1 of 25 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

PR #1689, opened 2026-08-06
docs: fix doubled words in oapi endpoint and operation docs

latent-9:
Two small doubled-word typos in the oapi docs:

- `crates/oapi/docs/endpoint.md` (included into the `#[endpoint]` macro docs via `include_str!`): "Value can be be [`str`]" -> "Value can be [`str`]" (4 occurrences).
- `crates/oapi/src/openapi/operation.rs`: "can be used for for the operation" -> "can be used for the operation".

Documentation only.


latent-9:
shanksyou legend

PR #1691, opened 2026-08-07
Add RFC 9457 problem details support

chrislearn:
## Summary

- add the feature-gated generic `Problem<Extensions = NoExtensions>` RFC 9457 response type, using Rust `kind` for the JSON `type` member
- provide `PlainProblem`, typed extensions via `with_extensions`, and a dynamic extension builder
- add `ProblemGoal` for opt-in catcher rendering and safe `StatusError` conversion
- compose typed extension schemas into Problem responses with `salvo-oapi`

## Testing

- `cargo +nightly fmt --all -- --check`
- `cargo check --all --bins --examples --tests`
- `cargo check -p salvo --no-default-features --features rfc9457,oapi`
- `cargo test -p salvo_core --all-features --lib`
- `cargo test -p salvo-oapi --all-features --lib`
- `cargo test --workspace --doc`
- targeted no-default-feature checks and RFC 9457 unit/integration tests
- targeted Clippy checks for `salvo_core` and `salvo-oapi` with the `rfc9457` feature

Closes #1690

PR #1692, opened 2026-08-10
Fix closed problem extension schemas

chrislearn:
## Summary

- merge object extension schemas into the RFC 9457 problem schema instead of composing them with `allOf`
- preserve closed-object behavior such as `additionalProperties: false` while allowing the standard problem members
- add a regression test using `#[serde(deny_unknown_fields)]`

## Why

The `allOf` composition introduced in #1691 makes a closed extension schema validate the standard `type`, `title`, and `status` members as unknown properties. As a result, generated clients and JSON Schema validators can reject valid `Problem<Extensions>` payloads.

## Impact

Typed problem extensions now produce a single merged object schema when the extension resolves to an object. Non-object schemas retain the existing `allOf` fallback.

## Validation

- `cargo test -p salvo-oapi --features "rfc9457,non-strict-integers"`
- `cargo clippy -p salvo-oapi --features rfc9457 --tests --no-deps`

### DIFF
--- a/crates/oapi/src/lib.rs
+++ b/crates/oapi/src/lib.rs
@@ -802,6 +802,37 @@ fn problem_base_schema(components: &mut Components) -> RefOr<schema::Schema> {
 }
 
 #[cfg(feature = "rfc9457")]
+fn problem_schema_with_extensions(
+    components: &Components,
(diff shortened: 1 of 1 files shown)


chrislearn:
@codex review

PR #1693, opened 2026-08-20
Stop serving XML-based files inline from static handlers

chrislearn:
## Summary

Fixes a stored XSS reported by **sl91994** (CWE-79) affecting `StaticDir` / `StaticFile` and the underlying `salvo_core::fs::NamedFile`.

`build_content_disposition` classified files by their **top-level** media type:

```rust
match (content_type.type_(), content_type.subtype()) {
    (mime::IMAGE | mime::TEXT | mime::VIDEO | mime::AUDIO, _)
    | (_, mime::JAVASCRIPT | mime::JSON) => "inline",
    _ => "attachment",
}
```

So `image/svg+xml` matched on `image` and `text/xml` matched on `text`, and both were served `inline`. Both are *active* content:

- an SVG carries `<script>` elements and event-handler attributes;
- any XML document can name an XSLT stylesheet via `<?xml-stylesheet ?>` and render scripted HTML from it.

An app serving user uploads through `StaticDir` therefore turned an uploaded "image" into stored XSS in its own origin. Extension/content-type allowlists don't help — `.svg` with `Content-Type: image/svg+xml` *is* a legitimate image by every check an upload endpoint typically makes. And neither `StaticDir` nor `StaticFile` exposed any way to override the disposition (`grep -rn "disposition" crates/serve-static/src/` returned nothing).

## The fix

The classification was **already** fail-safe for `application/xml` and `application/xhtml+xml` — they reach `attachment` because their top-level type isn't in the inline list. The bug is that the same documents spelled `image/svg+xml` or `text/xml` slip past. This extends the existing behavior to any `xml` subtype or `+xml` suffix, so the XML family is handled consistently:

| Content-Type | before | after |
|---|---|---|
| `image/svg+xml` | `inline` | `attachment` |
| `text/xml`, `text/xsl` | `inline` | `attachment` |
| `application/xml`, `application/xhtml+xml` | `attachment` | `attachment` |
| `image/png`, `text/plain`, `video/mp4` | `inline` | `inline` |
| `text/html` | `inline` | `inline` |

`text/html` deliberately keeps defaulting to `inline` — serving HTML documents is what a static file server is for. The vector this closes is the one where a *non-HTML* type sneaks past upload validation.

Additionally:

- `NamedFile` and `StaticEmbed` responses now send `X-Content-Type-Options: nosniff` (the report noted no `nosniff` anywhere in the tree).
- `StaticDir` gains `disposition_type` and `use_content_type_options`; `StaticFile` gains `disposition_type`, `attached_name` and `use_content_type_options` — so directories of trusted assets can opt back into inline rendering, which previously had no public API at all.

## Compatibility

Serving an SVG or XML file now sends `Content-Disposition: attachment`, so **following a link to one downloads it** instead of rendering, and `<object>` / `<embed>` / `<iframe>` no longer display it. Referencing the same file from ``, CSS `url()` or `<use>` is **unaffected** — browsers ignore `Content-Disposition` on subresource loads, so SVG icons and logos keep working.

Sites that intentionally serve trusted SVGs for direct viewing restore the old behavior with:

```rust
StaticDir::new(["assets"]).disposition_type("inline")
```

`StaticDir` is `#[non_exhaustive]`, so the two new public fields are not a breaking change.

## Verification

Six tests added. I confirmed they actually reproduce the vulnerability by stubbing `is_scriptable_xml` to `false` (restoring the old classification) and re-running:

```
thread 'tests::test_static_dir_does_not_serve_uploaded_svg_inline' panicked at
crates\serve-static\src\lib.rs:226:13:
poc.svg served with `inline`
```

With the fix in place: 345 tests pass across `salvo_core` and `salvo-serve-static`, `cargo check --workspace --all-features` is clean, `cargo fmt --check` is clean, and `cargo clippy` reports no new warnings (the 4 existing ones are all in untouched files).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   when an application served attacker-supplied uploads. `image/svg+xml` and `text/xml` were
   classified as `inline` by their top-level `image`/`text` type, so an uploaded SVG carrying
   `<script>` — or an XML document naming an XSLT stylesheet through `<?xml-stylesheet ?>` —
-  executed script in the serving origin when opened. Any content type with an `xml` subtype or
-  an `+xml` suffix now defaults to `attachment`, matching how `application/xml` and
-  `application/xhtml+xml` already behaved. Reported by sl91994.
+  executed script in the serving origin when opened. XML-based content types now default to
+  `attachment`, matching how `application/xml` and `application/xhtml+xml` already behaved.
+  That covers the `+xml` suffix, an `xml` subtype, RFC 7303's `xml-dtd` and
+  `xml-external-parsed-entity`, and the legacy `text/xsl` that `<?xml-stylesheet ?>` itself
+  names. `text/html` still defaults to `inline`. Reported by sl91994.
 - `NamedFile` and `StaticEmbed` responses now carry `X-Content-Type-Options: nosniff`.
 
 ### Added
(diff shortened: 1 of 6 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

PR #1694, opened 2026-08-20
Advertise the gzip coding implied by a .svgz extension

chrislearn:
## Summary

A `.svgz` is a gzipped SVG. Its extension names both a media type *and* the coding applied to it, but an extension resolves to a **single** media type, so `mime_infer` reports the type of the document *inside* the stream:

```rust
("svgz", &["image/svg+xml"]),
```

Nothing recovered the coding, so the response described a gzip stream as `image/svg+xml` and no client could render it. `StaticDir` only ever sets `Content-Encoding` from its precompressed **sidecar** lookup (`logo.svg.gz`), and `svgz` isn't in `compressed_variations`, so that path never fired either.

Reproduced before the fix — the `Content-Encoding` assertion is what fails:

```
thread 'tests::test_serve_static_dir_marks_svgz_as_gzip_encoded' panicked at
crates\serve-static\src\lib.rs:221:61
```

## The fix

The coding is inferred from the extension in `NamedFileBuilder::build`, right next to the media type inference it belongs with. That placement matters: the bug is not specific to the directory handler — `NamedFile::open("logo.svgz")` and `StaticFile::new("logo.svgz")` go through the same builder and were equally broken. Fixing it at the choke point covers all three rather than just `StaticDir`.

An explicitly configured encoding still wins (`content_encoding.or_else(...)`), which keeps `StaticDir`'s sidecar lookup authoritative for the files it picks.

X3D's compressed interchange forms are the same shape and get the same treatment:

| Extension | Content-Type | Content-Encoding |
|---|---|---|
| `.svgz` | `image/svg+xml` | `gzip` |
| `.x3dz` | `model/x3d+xml` | `gzip` |
| `.x3dvz` | `model/x3d+vrml` | `gzip` |
| `.x3dbz` | `model/x3d+binary` | `gzip` |
| `.gz`, `.tgz` | *(unchanged)* | *(none)* |

`.gz` and `.tgz` are deliberately excluded: there the gzip stream **is** the representation being served, not a coding applied to something else, so they must keep their own content type and arrive undecoded.

## A caveat worth recording

The coding is advertised unconditionally rather than negotiated. A client sending `Accept-Encoding: identity` is still handed gzip, which is not strictly conformant. The alternative would be decompressing on the fly or refusing the request, and the file on disk simply is gzip — this is also what `AddEncoding gzip svgz` does on Apache. Absent `Accept-Encoding` is fine either way, since RFC 9110 §12.5.3 treats a missing header as "any coding is acceptable".

## Relationship to #1693

Independent, and verified to merge cleanly in either order. I ran a trial merge of the two branches: no conflicts, and 349 tests pass on the merged result. The two changes compose the way you would want on the file that touches both:

```
content-type        = image/svg+xml
content-encoding    = gzip           <- this PR
content-disposition = attachment     <- #1693
x-content-type-options = nosniff     <- #1693
```

## Verification

Four tests added, covering the `StaticDir` response headers and body pass-through, the extension table (including that `.gz`/`.tgz` are excluded and that matching is case-insensitive, following `mime_infer`), and that an explicit `content_encoding` overrides the inferred one. Full suites pass for `salvo_core` and `salvo-serve-static`, `cargo check --workspace --all-features` is clean, `cargo fmt --check` is clean, and clippy reports no new warnings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ### Fixed
 
+- `.svgz` files are now served with `Content-Encoding: gzip`. Their extension names both a
+  media type and the coding applied to it, but an extension resolves to a single media type,
+  so the response described the gzip stream as `image/svg+xml` and no client could render it.
+  The gzipped X3D forms `.x3dz`, `.x3dvz` and `.x3dbz` are handled the same way. `.gz` and
+  `.tgz` are unaffected, since there the gzip stream is the representation being served.
 - `PathItem` no longer duplicates operations into `extensions` when deserialized, which
   previously made a parsed path item re-serialize with repeated keys.
 - OAuth2 flows are now resolved by their flow name instead of by shape, so a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 - `Content::from_ref` / `Content::ref_location`, so a `content` map entry can reference a
   reusable Media Type Object without changing those maps to `RefOr<Content>`.
 - `examples/oapi-3-2` demonstrates emitting a 3.2 document with a `QUERY` route.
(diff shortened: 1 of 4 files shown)


chrislearn:
@codex review

PR #1697, opened 2026-08-21
Support JSON values in cookie extractors

### LINKED ISSUE #1695: How to let json formatted struct extracted from a cookie
As I know, for now if I want an aggregate struct and store it in cookie as json, I can either read it as a string a parse it in handler which lost opai support in endpoint, or manually implement Extractible for my struct.

An example is like this
```rust
#[derive(Serialize, Deserialize, ToSchema, Clone, Debug, Extractible)]
pub struct AuthToken {
    pub pkce_verifier: String,
    pub csrf_token: String,
    pub nonce: String,
    pub from_url: Option<String>,
}

#[handler]
async fn callback(c: CookieParam<AuthToken,true>){
}
```

And let it to use macro to extract
```rust
#[derive(Serialize, Deserialize, Extractible, Debug)]
#[salvo(extract(default_source(from = "query")))]
pub struct CBParam {
    #[salvo(extract(source(from = "cookie", parse = "json")))]
    pub tok: AuthToken,
}

```
And by the way, the skills.md says there is a cookie source, but actually doesn't exist.

chrislearn:
## Summary
- accept `cookie` as an `Extractible` field source and honor explicit JSON parsing
- let `CookieParam<T>` fall back to JSON for structured cookie values while preserving scalar conversion
- add encoded-cookie regression coverage and document the new behavior

## Testing
- `cargo test -p salvo_core --all-features`
- `cargo check -p salvo_core --no-default-features`
- `cargo test -p salvo_macros`
- `cargo test -p salvo-oapi --all-features`
- Clippy on the affected crates with warnings denied after allowing the repository's existing Rust 1.97 lints

Fixes #1695

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   into inline rendering, which previously had no public API on either handler.
 - `NamedFileBuilder::disposition_name`, which sets the name `Content-Disposition` reports
   without forcing the disposition to `attachment` the way `attached_name` does.
+- Cookie extractors now support structured JSON values. `CookieParam<T>` falls back to JSON when
+  scalar conversion fails, while `#[derive(Extractible)]` accepts explicit cookie field sources
+  such as `#[salvo(extract(source(from = "cookie", parse = "json")))]`.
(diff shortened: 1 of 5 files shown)


chrislearn:
@codex review

chrislearn:
@codex review

